PR #23, opened 2024-12-16
Analysis downloads/qc/taxa

SandyRogers:
This PR:
- adds a `quality_control` field to `Analysis` object.
  - This is some JSON. It should really have a Pydantic schema for it – this is TODO I'm afraid.
  - adds a minimal importer for Amplicon V6 outputs that takes the `.summary` json path from the `*.fastp.json` file and puts it into the DB. This includes metadata like total bases and GC content before and after filtering.
  - new field is available on analysis detail API endpoint as `quality_control_summary`.
- improves how taxonomy files are imported, by using a few new `FileRule`s and making `DownloadFile`s on the `Analysis` object when an Amplicon V6 is finished.
  - TODO: this is just minimal for now... it adds a Krona plot file as the download, not all the other ones. Proof of concept.
  - TODO: there is no file serving currently in the k8s config; and there is also no datamover job to move files onto a public (transfer services) location to be served by external HTTP engines. 
- adds DADA2 results to possible `TaxonomySources`.
  - however these require different handling due to file names including variable region info. Currently code has some warnings/todos/NotImplementedErrors about this.

- adds some small helpers to admin panels for better navigating `JSONField`s in the various common flavours we have.


workflows/data_io_utils/mgnify_v6_utils/amplicon.py
```diff
@@ -0,0 +1,199 @@
...
+        )
+
+    if not qc_dir.path.is_dir():
+        print(f"No qc dir at {qc_dir.path}. Nothing to import.")
+        return
+
+    # TODO rules and downloadfile for multiqc
+    qc_dir.files.append(
+        File(
+            path=qc_dir.path / f"{analysis.run.first_accession}_multiqc_report.tsv",
```
KateSakharova, line 165: Should those filenames be also in config or somewhere else in one place defined? 
I think for tools like fastp or multiqc that we will use in many/all our pipelines we can re-use names of files (that will help also developing other pipelines with the same filenames/structures).

KateSakharova (approved):
Checked PR 👍  just left one tiny question

workflows/data_io_utils/mgnify_v6_utils/amplicon.py
```diff
@@ -0,0 +1,199 @@
...
+    analyses.models.Analysis.TaxonomySources.DADA2_PR2: "DADA2-PR2",
+    analyses.models.Analysis.TaxonomySources.DADA2_SILVA: "DADA2-SILVA",
+}
+
+
+def import_taxonomy(
+    analysis: analyses.models.Analysis,
+    dir_for_analysis: Path,
+    source: analyses.models.Analysis.TaxonomySources,
+    allow_non_exist: bool = True,
```
mberacochea, line 47: Maybe, `allow_missing`?

mberacochea:
Great stuff @SandyRogers. I think this is going in the right direction, but I feel we should simplify the code a bit more (not sure how... I will mull it over during the break).

mberacochea:
Onward... I'm merging this!

PR #24, opened 2024-12-17
added pre-qc-failed status filter

KateSakharova:
Added filtering for not re-trying pre-qc failed assemlies and amplicons

### DIFF
--- a/analyses/models.py
+++ b/analyses/models.py
@@ -267,13 +267,13 @@ class Assembly(TimeStampedModel, ENADerivedModel):
         def default_status(cls):
             return {
                 cls.ASSEMBLY_STARTED: False,
+                cls.PRE_ASSEMBLY_QC_FAILED: False,
                 cls.ASSEMBLY_FAILED: False,
                 cls.ASSEMBLY_COMPLETED: False,
                 cls.ASSEMBLY_BLOCKED: False,
                 cls.ASSEMBLY_UPLOADED: False,
                 cls.ASSEMBLY_UPLOAD_FAILED: False,
                 cls.ASSEMBLY_UPLOAD_BLOCKED: False,
-                cls.PRE_ASSEMBLY_QC_FAILED: False,
             }
 
     status = models.JSONField(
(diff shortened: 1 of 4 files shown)


analyses/models.py
```diff
@@ -272,6 +272,7 @@ def default_status(cls):
                 cls.ASSEMBLY_UPLOADED: False,
                 cls.ASSEMBLY_UPLOAD_FAILED: False,
                 cls.ASSEMBLY_UPLOAD_BLOCKED: False,
+                cls.PRE_ASSEMBLY_QC_FAILED: False,
```
mberacochea, line 275: Pleae put "cls.PRE_ASSEMBLY_QC_FAILED: False," after "cls.ASSEMBLY_STARTED: False," :)

analyses/models.py
```diff
@@ -504,7 +505,7 @@ class AnalysisStates(str, Enum):
         ANALYSIS_COMPLETED = "analysis_completed"
         ANALYSIS_BLOCKED = "analysis_blocked"
         ANALYSIS_FAILED = "analysis_failed"
-        ANALYSIS_QC_FAILED = "analysis_qc_failed"
+        ANALYSIS_PRE_QC_FAILED = "analysis_pre_qc_failed"
```
mberacochea, line 508: For analyses we don't have a "pre-qc".. it's just QC, right?

workflows/flows/assemble_study.py
```diff
@@ -126,6 +126,7 @@ def get_assemblies_to_attempt(study: analyses.models.Study) -> List[Union[str, i
         **{
             f"status__{analyses.models.Assembly.AssemblyStates.ASSEMBLY_COMPLETED}": False,
             f"status__{analyses.models.Assembly.AssemblyStates.ASSEMBLY_BLOCKED}": False,
+            f"status__{analyses.models.Assembly.AssemblyStates.PRE_ASSEMBLY_QC_FAILED}": False,
```
mberacochea, line 129: This one should go first; it doesn't change how it works—just improves readability

mberacochea (changes requested):
Good catch.. not sure if we should rename the analyses QC to PRE-QC

SandyRogers (approved):

KateSakharova:
@SandyRogers thank you, I will rebase from your branch!

PR #26, opened 2024-12-19
Resolved Schema validation errors on annotations API

MGS-sails:
the MGnifyAnalysisTypedAnnotation schema inherits from MGnifyAnalysisDetail which has certain required fields in it's response structure. 
This pull request adds the missing fields, thereby addressing the previous error appearing on the API


mberacochea (approved):
LGTM :). Thanks!

PR #27, opened 2025-01-06
Private data surface

MGS-sails:
1. Implemented exclusion of private data by default on Analyses and Studies endpoints
2. Added basis for exposing private data only for future my-data endpoints
3. the my-data endpoints need to have auth protection and a check for data ownership
4. Added privacy filters for Django admin related views
5. Adjusted related model files to allow for exposing private/public data only as required.


emgapiv2/api.py
```diff
@@ -237,6 +237,37 @@ def list_mgnify_analyses(request):
...
+@api.get(
+    "/my-data/studies",
+    response=List[MGnifyStudy],
+    tags=[ApiSections.STUDIES.value],
+    summary="List all private studies available from MGnify",
+    operation_id="list_private_mgnify_studies",
+)
+def list_private_mgnify_studies(request):
+    qs = analyses.models.Study.objects.private_only()
+    return qs
```
SandyRogers, line 268: We need an ownership check here, else it lets anybody list all private objects.
I think we want an extra manager method on the privacy manager base, like `def owned_only(webin): qs = qs.filter(webin_submitter__iexact=webin)`, and then here we can do `qs = analyses.models.Study.objects.owned_only(request.user)` (assuming we also add an auth layer)

emgapiv2/test_data_privacy.py
```diff
@@ -0,0 +1,140 @@
...
+@pytest.fixture
+def admin_client():
+    from django.test import Client
+
+    admin_user = User.objects.create_superuser(
+        username="admin", email="admin@example.com", password="adminpass123"
+    )
+    client = Client()
+    client.force_login(admin_user)
+    return client
```
SandyRogers, line 140: There should already be an `api_client` and `admin_client` fixture provided by `pytest-django` automatically - we shouldn't need to define these ones.

requirements.txt
```diff
@@ -32,3 +32,7 @@ requests~=2.32.3
 httpx~=0.27.2
 typing_extensions~=4.12.2
 asgiref~=3.8.1
+
+pytest~=8.3.4
+djangorestframework~=3.15.2
+responses~=0.25.3
```
SandyRogers, line 38: pytest and responses are already provided by `requirements-dev.txt` – we should leave them there since they aren't needed by the production build. `djangorestframework` can also be removed I think, if we use the `pytest-django` provided fixtures for `api_client/admin`.

SandyRogers (changes requested):
Thanks @MGS-sails ! We are well on our way – see comments inline.

emgapiv2/api.py
MGS-sails: Thanks @SandyRogers 
Actually.. I left this as is on purpose. 
you're definitely right that we need an ownership check / some sort of middleware here. 
However, I just put this as sort of a placeholder. Currently I just used it to double check that private/public data are indeed being filtered.
And yes, my plan was also to add the auth layer in the future.

analyses/tests.py
```diff
@@ -196,3 +203,125 @@ def test_analysis_inheritance(
...
+# @pytest.fixture
+# def admin_client():
+#     from django.test import Client
+#
+#     admin_user = User.objects.create_superuser(
+#         username="admin", email="admin@example.com", password="adminpass123"
+#     )
+#     client = Client()
+#     client.force_login(admin_user)
+#     return client
```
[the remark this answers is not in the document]
MGS-sails: this was debris from when I moved the tests to a separate file. 
this is now removed

requirements.txt
MGS-sails: I've removed to other dependencies, but have kept djangorestframework. 
From my observations, pytest-django does not seem to have an api_client fixture. 
So I've maintained my implementation of that. but have removed the admin-client

emgapiv2/test_data_privacy.py
MGS-sails: found admin-client, and thus removed my implementation.
However, I  could not find an api-client fixture out of the box. So I've maintained that

analyses/models.py
```diff
@@ -64,6 +64,26 @@ def lineage_to_path(lineage: str) -> str:
         return re.sub(r"[^a-zA-Z0-9._]", "", underscore_punctuated)
 
 
+class BasePrivacyFilterManager:
```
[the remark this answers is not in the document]
MGS-sails: This has been implemented

emgapiv2/test_data_privacy.py
```diff
@@ -0,0 +1,124 @@
...
+    assert analysis.annotations == {"test": "data"}
+    all_analyses = Analysis.all_objects_and_annotations.all()
+    assert len(all_analyses) == 2
+
+
+@pytest.fixture
+def api_client():
+    from rest_framework.test import APIClient
+
+    return APIClient()
```
SandyRogers, line 124: I think this fixture is already provided by pytest-django isn't it?

requirements.txt
```diff
@@ -32,3 +32,5 @@ requests~=2.32.3
 httpx~=0.27.2
 typing_extensions~=4.12.2
 asgiref~=3.8.1
+
+djangorestframework~=3.15.2
```
SandyRogers, line 36: If we can use the `api_client` fixture from pytest-django, we shouldn't need this I believe.
```suggestion
```

SandyRogers (commented):
Thanks @MGS-sails ! Just a few last tidy-up comments inline, then this is good to merge.

emgapiv2/test_data_privacy.py
MGS-sails: HI @SandyRogers I checked for this and it doesn't seem to be provided. 
I removed the one for the AdminClient fixture however.

requirements.txt
MGS-sails: @SandyRogers please see my comment here:

https://github.com/EBI-Metagenomics/emgapi-v2/pull/27#discussion_r1907131347

emgapiv2/test_data_privacy.py
SandyRogers: Ah yes, sorry I misremembered. The one we are already using is `ninja_api_client`.

MGS-sails: @SandyRogers I've replaced this with the ninja_api_client

SandyRogers (approved):

PR #29, opened 2025-01-17
Cluster Job store

SandyRogers:
This PR:
- Refactors the "slurm flow" logic, that runs HPC workloads on the slurm cluster instead of on the prefect agent
  - The crucial change is that we now keep an in-database store of the submitted slurm jobs (`OrchestratedClusterJob`s). This seems redundant (given prefect's task DB, slurm's job DB, and even nextflow's platform DB), but it gives more reliable and explicit control over how to **reconnect** to slurm jobs (e.g. when prefect flow is restarted during a nextflow pipeline run) and how to **resubmit** jobs (e.g. when a pipeline failed and we want to run it again with no obvious change in inputs).
  - The other big change is introducing `SlurmResubmitPolicy`s as a way to control this logic in different scenarios. For example there is a `ResubmitIfFailedPolicy` (basically: run a new slurm job if we ran exactly the same thing before and it failed) and a `ResubmitAlwaysPolicy` (basically: even if we successfully ran an identical job before, still run a new one. Useful e.g. for a job that is a trigger.).
- Adds a `JSONFieldWithSchema` django field. 
  - This is a JSONField, where the content must match a Pydantic model (or a list of a model). This is useful for storing the slurm job parameters that were submitted (so `OrchestratedClusterJob.job_submit_description` is a JSON field that must follow a `SlurmJobSubmitDescription` schema. 
  - This is handy for de/serialization purposes, since you don't have to rely on knowing JSON keys as strings, but also means we can more reliably put database indexes on the JSON content.
    - Having indexes means it is fast to search the `OrchestratedClusterJob` table for previously submitted identical jobs, because we can do a lookup on the actual job script.
  - This is also being used to keep track of hashes of the designated input files to cluster jobs (i.e. nextflow samplesheets) so that `SlurmResubmitPolicy` can specify whether a change in input file content should count as a different job or not (usually: yes).
- Cluster logs (i.e. the slurm stdout) and Nextflow trace (i.e. the structure resource utilisation per task) are stored in the `OrchestratedClusterJob`. The trace is a TSV file, but this is parsed into JSON for the DB when a job finishes successfully (so that it could be indexed/queried later, e.g. for usage analytics or for determining new `ComputeResourceHeuristic` values).

# Visual changes:
This is the django admin panel showing the jobs that flows have launched on Slurm:

Screenshot 2025-01-17 at 19 23 25

This is the django admin panel showing the cluster log (just text) and nextflow trace (converted to JSON for future postgres indexing/querying):

Screenshot 2025-01-20 at 16 57 49


emgapiv2/model_utils.py
```diff
@@ -0,0 +1,110 @@
...
+
+        class SchemaDecoder(_PydanticDecoder):
+            def __init__(inner_self, *args, **kwargs):
+                super().__init__(schema=schema, is_list=is_list, *args, **kwargs)
+
+        return SchemaDecoder
+
+    def deconstruct(self):
+        """
+        Ensure the 'decoder' argument is excluded during migration serialization.
```
mberacochea, line 81: Deep down in the guts of django :)

workflows/migrations/0005_alter_orchestratedclusterjob_input_files_hashes_and_more.py
```diff
@@ -0,0 +1,34 @@
...
+        ("workflows", "0004_orchestrated_cluster_jobs"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="orchestratedclusterjob",
+            name="input_files_hashes",
+            field=emgapiv2.model_utils.JSONFieldWithSchema(
+                is_list=True,
+                schema=workflows.models.OrchestratedClusterJob.JobInputFile,
```
mberacochea, line 21: Nice :+1:

emgapiv2/model_utils.py
SandyRogers: Yeh :) It stops django making broken migrations due to unserializable / unreferencable code. It would be possible to fix so that the migration files include a more meaningful spec for the schema, but I didn't think it was worth the extra work.

workflows/models.py
```diff
@@ -1,3 +1,170 @@
...
+        """
+
+        model_config = ConfigDict(extra="allow")
+
+        name: str = Field(..., description="Name of the job")
+        script: str = Field(..., description="Content of the job script/command")
+        memory_per_node: Optional[Union[str, int]] = Field(
+            None, description="Memory required for the job"
+        )
+        time_limit: Optional[str] = Field(
```
mberacochea, line 82: Should this one be datetime?

workflows/models.py
```diff
@@ -1,3 +1,170 @@
...
+    class SlurmJobSubmitDescription(BaseModel):
+        """
+        Pydantic schema for storing a slurm JobSubmitDescription object as json.
+        """
+
+        model_config = ConfigDict(extra="allow")
+
+        name: str = Field(..., description="Name of the job")
+        script: str = Field(..., description="Content of the job script/command")
+        memory_per_node: Optional[Union[str, int]] = Field(
```
mberacochea, line 79: Should we normalize this here? Either to bytes, or something like Nextflow (1.GB) kind of thing (in the future not now BTW)

workflows/models.py
SandyRogers: The idea here was to model the params that `pyslurm` takes (so same format as slurm CLI... and *hopefully* same as Slurm REST API – I need to look into that at some point). So `time_limit` is a string of format like `4-12:00:00` for 4.5 days. I thought about making a custom type for this based on `timedelta`, but decided it was slightly unneeded polish for now.

workflows/models.py
SandyRogers: Yeh good idea, would be nice to avoid potential bugs by doing so. This is currently (like `time_limit`) just a representation of `pyslurm` params. I think the interpretation of ints here may be configurable in slurm config though, which is a recipe for weird future bugs...

workflows/models.py
mberacochea: agreed

workflows/prefect_utils/slurm_flow.py
```diff
@@ -392,10 +352,22 @@ def compute_hash_of_input_file(
     return input_files_hash.hexdigest()
 
 
+@task(log_prints=True)
+def store_nextflow_trace(orchestrated_cluster_job: OrchestratedClusterJob):
+    job_description: OrchestratedClusterJob.SlurmJobSubmitDescription = (
```
mberacochea, line 357: Should we have a flag to raise an error if no trace is found? This data is quite important, and I would like to avoid losing it

workflows/tests/test_cluster_jobs_integration.py
```diff
@@ -102,39 +131,64 @@ async def test_run_cluster_job_state_persistence(
...
         memory="100M",
         environment={},
+        resubmit_policy=ResubmitIfFailedPolicy,
+        working_dir=tmp_path,
         input_files_to_hash=[tmp_path / "my_inputs.csv"],
     )
-    assert job_id_initial_with_hash != job_id_repeat_with_hash
+    assert job_initial_with_hash.cluster_job_id != job_repeat_with_hash.cluster_job_id
+
+    # if input file changes, but policy ignores that, should NOT be new job
```
mberacochea, line 174: I'm struggling to think of a case when a job that has a modified input should not be triggered

mberacochea (commented):
This is quite a lot of work :factory_worker:.

I've left a few comments, I'm happy to chat about them.

workflows/tests/test_cluster_jobs_integration.py
SandyRogers: I think it would be very niche, yes, and only really added it because I knew it'd be easier to do so with the architecture fresh in mind. Maybe premature over-engineering. A use case would be if we know some input file is non-deterministic, but we don't really mind:

```python
@task(cached=False)
def get_things_from_very_fast_api(study='ERP1') -> Path:
    csv = internet.get(f"my-api.org/{study}).as_csv(f"{study}_data.csv")
    return csv.path

csv_path = get_things_from_very_fast_api(study)

do_expensive_stuff_with_data(
    study,
    csv_path,
    resubmit_policy=DontResubmitIfOnlyInputFilesChangePolicy
)  # <-- if this had previously succeeded, we may not care that the csv is now in a different order

```

workflows/prefect_utils/datamovers.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    expected_time = kwargs.pop("expected_time", timedelta(hours=2))
+    memory = kwargs.pop("memory", "1G")
+
+    if not "environment" in kwargs:
+        kwargs["environment"] = {}
+
+    return await run_cluster_job(
+        name=f"Move {file_path_shortener(source)} to {file_path_shortener(target)}",
+        command=f"{move_command} {source} {target}",
```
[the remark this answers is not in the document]
SandyRogers: Good catch

workflows/tests/test_cluster_jobs_integration.py
mberacochea: perfect.. I was thinking something like that but couldn't come up with a "real" example of stuff we usually run. All good then, carry on :)

PR #31, opened 2025-01-23
Feature/amplicon results

MGS-sails:
Amplicon results workflow
    Created workflow to move amplicon pipeline results;
    made ClusterJobFailedException more verbose to aid in debugging


workflows/prefect_utils/slurm_flow.py
```diff
@@ -322,7 +322,18 @@ def cancel_cluster_job(name: str):
...
+        self.job_id = job_id
+        self.state = state
+        self.message = message
+        super().__init__(self._format_message())
+
+    def _format_message(self):
+        msg = f"Cluster job {self.job_id} failed with state {self.state}"
+        if self.message:
+            msg += f"\nDetails: {self.message}"
+        return msg
```
SandyRogers, line 336: nice

workflows/flows/copy_amplicon_pipeline_results.py
```diff
@@ -0,0 +1,15 @@
+from prefect import flow
+
+from workflows.prefect_utils.slurm_flow import move_data
+
+
+@flow(
+    name="Copy Amplicon Pipeline Results",
+    flow_run_name="Copy Amplicon Pipeline Results",
+)
+async def copy_amplicon_pipeline_results():
```
SandyRogers, line 10: This should take a param of `analysis: analyses.models.Analysis`, because in the end we will call this `copy_amplicon_pipeline_results` flow as a subflow of the `analysis_amplicon_study` flow.

So, this flow can get the `move_data` source dir from `analysis.results_dir` and `EMG_CONFIG.slurm.default_workdir`.

The destination (FTP) should be something like `Path(EMG_CONFIG.slurm.ftp_results_dir) / analysis.study.first_accession[:4] / analysis.study.first_accession`

SandyRogers (changes requested):
Thanks @MGS-sails . Have added some clarifications of what the copy flow needs to do, inline.

mberacochea (changes requested):
Thanks Mahfouz!

workflows/prefect_utils/slurm_flow.py
```diff
@@ -489,7 +500,17 @@ async def run_cluster_job(
             is_job_in_terminal_state = True
 
         if slurm_status_is_finished_unsuccessfully(job_state):
-            raise ClusterJobFailedException()
+            error_details = None
+            try:
+                job = pyslurm.db.Job(job_id).load(job_id)
+                job_log_path = Path(job.working_directory) / Path(f"slurm-{job_id}.out")
```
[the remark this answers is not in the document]
SandyRogers: Slurm (or at least the pyslurm object) doesn't have a property for the stdout file sadly.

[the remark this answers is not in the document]
mberacochea: _sad_

workflows/flows/copy_amplicon_pipeline_results.py
KateSakharova, line 10: according to RP #34 should it still be `async`?

KateSakharova (approved):
Nice!

workflows/flows/copy_amplicon_pipeline_results.py
MGS-sails: think I'll keep this as a normal function, unless the need for making it async arises

workflows/flows/copy_amplicon_pipeline_results.py
```diff
@@ -0,0 +1,121 @@
...
+    #     'taxonomy-summary': ['*.txt', '*.tsv', '*.mseq', '*.html'],
+    #     'sequence-categorisation': ['*.fa', '*.fasta', '*.deoverlapped'],
+    #     'qc': ['*.html', '*.json', '*.tsv', '*.gz'],
+    #     'primer-identification': ['*.tsv', '*.fasta', '*.json'],
+    #     'asv': ['*.tsv', '*.fasta'],
+    #     'amplified-region-inference': ['*.tsv', '*.txt'],
+    #     '.': ['sample.txt','study_multiqc_report.html', 'qc_*.csv', 'primer_*.json', 'manifest.json']
+    # }
+
+    ALLOWED_PATTERNS = {
```
mberacochea, line 58: Why can't we do all folder and file creation with python?

mberacochea (changes requested):
I'm probably missing something, but why is the copy_amplicon_pipeline_results.py looking so drafty?

docker-compose.yaml
```diff
@@ -6,7 +6,7 @@ include:
 
 services:
   prefect-database:
-    image: postgres:16
+    image: postgres:17
```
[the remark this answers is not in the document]
SandyRogers: Currently our EBI-managed DB is 14. But newly created ones are currently 16, so we're using that here anticipating that we will probably deploy a new production DB and want to match that. Django support currently goes back to v13 so I don't think v14 vs v16 vs v17 will matter currently.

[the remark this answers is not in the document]
MGS-sails: On my local I use version 17. 
I've changed it back to 16 for prod.

workflows/flows/copy_amplicon_pipeline_results.py
MGS-sails: @mberacochea I've changed how this works please have a look.
this part of the code does not actually move the files. 
It just checks for invalid files and then calls the move_data utility. 
which in turn triggers a slurm job.

SandyRogers (approved):
Thanks Mahfouz! Happy to merge this one now – I'll write it into the amplicon flow as I expand the file importer bit on that.

mberacochea (approved):
One small question

PR #34, opened 2025-01-31
Makes run_cluster_job a sync method

SandyRogers:
`run_cluster_job` was previously an `async` method, so that we could do `await run_cluster_job(...)`. This is not (any more) necessary, it can be a sync method to make it easier to bring into non-async flows.

This PR converts it from async to sync.


KateSakharova (approved):
Ahhh, back to normal life! Super :)

PR #38, opened 2025-02-07
Host reference genome

SandyRogers:
This PR:
- fetches host taxonomy ID for runs when pulling them from ENA
- adds a lookup of host-tax-id to known reference genomes, based on currently available options in miassembler
- adds the `--reference_genome` flag to miassembler calls, if one can be determined

The reference genome selection is based on an exact match of the NCBI tax ID, between the reference genomes we currently support and the ENA metadata on any run in a study. This isn't perfect, because host tax IDs might be missing in metadata or be specific at some higher rank. But it is a start. I considered adding a flow-suspend if the assemblies about to be launched were from a host-associated biome but a tax ID couldn't be found... but I think this is too common. A better bet would probably be a way to bulk update the runs of a study, in the DB, e.g. to assert a host TAX ID onto them.

Also I decided to keep the lookup of tax-id > reference genome as in-code, instead of in-db, since these are also an in-code Enum in miassembler so a code change seemed appropriate to add new ones.

### DIFF
--- a/workflows/flows/assemble_study_tasks/assemble_samplesheets.py
+++ b/workflows/flows/assemble_study_tasks/assemble_samplesheets.py
@@ -7,6 +7,7 @@ from typing import Iterable, Optional
 import django
 import pandas as pd
 from django.conf import settings
+from django.db.models import Q
 from prefect import flow, get_run_logger, task
 
 django.setup()
@@ -28,7 +29,7 @@ from workflows.prefect_utils.slurm_policies import (
 EMG_CONFIG = settings.EMG_CONFIG
 
 
-HOST_TAX_IDS_TO_REFERENCE_GENOME = {
+HOST_TAXON_TO_REFERENCE_GENOME = {
     "9031": "chicken.fna",  # Gallus gallus
     "8030": "salmon.fna",  # Salmo salar
     "8049": "cod.fna",  #  Gadus morhua
@@ -41,6 +42,18 @@ HOST_TAX_IDS_TO_REFERENCE_GENOME = {
     "9940": "sheep.fna",  # Ovis aries
     "3847": "soybean.fna",  # Glycine max
     "7955": "zebrafish.fna",  #  Danio rerio
+    "Gallus gallus": "chicken.fna",
+    "Salmo salar": "salmon.fna",
+    "Gadus morhua": "cod.fna",
+    "Sus scrofa": "pig.fna",
+    "Bos taurus": "cow.fna",
+    "Mus musculus": "mouse.fna",
+    "Apis mellifera": "honeybee.fna",
+    "Rattus norvegicus": "rat.fna",
+    "Oncorhynchus mykiss": "rainbow_trout.fna",
+    "Ovis aries": "sheep.fna",
+    "Glycine max": "soybean.fna",
+    "Danio rerio": "zebrafish.fna",
 }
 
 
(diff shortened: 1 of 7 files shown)


SandyRogers:
> the host_scientific_name is the one that sits in the simplest of sample checklists in ENA. I am not sure if they automatically populate the taxid from that but I wonder if it would be safer? 
> Any then the key change to host_scientific_name. If you know taxis are automatically populated then it looks good to me!

Thanks @Vkale1 . I ran a few queries on ENA and MGnify DBs. It looks like you need to use both tax id and scientific name to get all available metadata for the hosts we currently consider (chickens etc). So I have added host-scientific-names to the lookup as well and we now try using both to determine a reference genome.

Vkale1 (approved):

workflows/flows/assemble_study_tasks/assemble_samplesheets.py
```diff
@@ -28,6 +29,72 @@
...
+    "Sus scrofa": "pig.fna",
+    "Bos taurus": "cow.fna",
+    "Mus musculus": "mouse.fna",
+    "Apis mellifera": "honeybee.fna",
+    "Rattus norvegicus": "rat.fna",
+    "Oncorhynchus mykiss": "rainbow_trout.fna",
+    "Ovis aries": "sheep.fna",
+    "Glycine max": "soybean.fna",
+    "Danio rerio": "zebrafish.fna",
+}
```
KateSakharova, line 57: Should it be in config?

KateSakharova (approved):
I feel like prefect needs to be more flexible with possible input arguments. Like if user wants to add another host (maybe it is not set correctly in metadata, whatever), the same for assembler choice, for memory, etc... 
That is my conservative impression. I felt previous automation quity handy if you want to include another experiment type or another library layout. Just couple of args and woohoo. Here you will only see failed flow and cry for some time on that :)

workflows/flows/assemble_study_tasks/assemble_samplesheets.py
```diff
@@ -28,6 +29,72 @@
...
+    "8049": "cod.fna",  #  Gadus morhua
+    "9823": "pig.fna",  # Sus scrofa
+    "9913": "cow.fna",  # Bos taurus
+    "10090": "mouse.fna",  # Mus musculus
+    "7460": "honeybee.fna",  # Apis mellifera
+    "10116": "rat.fna",  # Rattus norvegicus
+    "8022": "rainbow_trout.fna",  # Oncorhynchus mykiss
+    "9940": "sheep.fna",  # Ovis aries
+    "3847": "soybean.fna",  # Glycine max
+    "7955": "zebrafish.fna",  #  Danio rerio
```
mberacochea, line 44: We need to re-oganize these files :(. I was aiming of having something like this https://ewels.github.io/AWS-iGenomes/ but in our FTP|hps.

workflows/flows/assemble_study_tasks/assemble_samplesheets.py
SandyRogers: It's a good point Kate. My sense was that since it was a coded Enum in miassembler, it should probably be one in automation too. I think config can be **over**used for structures like this... e.g. if we put this in config, and then want to add one more reference genome, we have to duplicate this entire dict, plus one line, into a config file to override it here (or use a fancier data structure) Hmm

workflows/flows/assemble_study_tasks/assemble_samplesheets.py
SandyRogers: Cool, intriguing. Is that something near-term @mberacochea (i.e. should I keep this PR open and wait for it) or one to refactor later?

mberacochea: I don't think so... I was leaving those thoughts here for posterity. There is a JIRA task, but we won't be working on this in the short term.

workflows/flows/assemble_study_tasks/assemble_samplesheets.py
SandyRogers: Yeh, that would be nicest way. Need to think about the best process for that, though 🤔

workflows/tests/test_assemble_study_flow.py
```diff
@@ -79,6 +80,8 @@ async def test_prefect_assemble_study_flow(
                 "library_strategy": "WGS",
                 "library_source": "METAGENOMIC",
                 "scientific_name": "metagenome",
+                "host_tax_id": "7460",
```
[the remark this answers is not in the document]
SandyRogers: I used strings, because if we use ints then we need to parse ENA API JSON strings to ints, catch parsing exceptions in case a non-int (`"host_tax_id": "unknown"`) somehow sneaks in there etc. I figured it was easier to just use strings and ignore the fact they happen to be ints in them.

workflows/flows/assemble_study_tasks/assemble_samplesheets.py
SandyRogers: Created task EMG-7399 to do this properly, thanks @KateSakharova and @mberacochea .

PR #41, opened 2025-02-13
Private assembly & prefect v3

SandyRogers:
This PR:
- adds support for assembly + upload of private ENA reads data, via brokering.
  - this is the main feature of the PR. Basic changes in `workflows/flows/assembly_study.py`.
  - assembly uploader flow/tasks are also updated, because for private data assemblies the `assembly_uploader` pushes assemblies into the reads study, rather than making a TPA study (`workflows/flows/upload_assembly.py`)
  - expands the use and syncing of "private" data types, by using `.is_private` and `.webin_submitter` fields over ~all models. There are lots more uses of the `.all_objects....` managers now instead of `.objects....`, since in flow code we typically want `all` (public AND private) data. 
- removes a lot of `async` code from flows/tasks/helpers/tests (should have been a separate PR, but things got mixed together).
- adds auth helpers for ENA (`workflows/ena_utils/ena_auth.py`) for easier management of webin/broker credentials.
- adds checkers for public/private/suppressed state of ENA studies based on combinations of unauthenticated and broken-authenticated API calls.
- upgrades prefect to v3 (and some other less important package bumps). This was needed to get some bugfixes in the prefect test harness, that aren't in v2.
  - this was mostly smooth, except for some changes needed to task concurrency config and fiddly changes to logging, because django/pytest/prefect all do complex things with swallowing logs. 
- adds ruff to pre-commit, to iron out unused imports amongst other things.


workflows/ena_utils/ena_api_requests.py
```diff
@@ -296,16 +315,38 @@ def create_ena_api_request(result_type, query, limit, fields, result_format="jso
...
     task_run_name="Get study from ENA: {accession}",
 )
-async def get_study_from_ena(accession: str, limit: int = 10) -> ena.models.Study:
+def get_study_from_ena(accession: str, limit: int = 10) -> ena.models.Study:
     logger = get_run_logger()
 
     logger.info(f"Will fetch from ENA Portal API Study {accession}")
 
+    is_public = is_ena_study_public(accession)
+    is_private = not is_public and is_ena_study_available_privately(accession)
```
mberacochea, line 336: Mayber we should rename _is_private_ with _privately_available_?

mberacochea (commented):
I didn't do a very thorough review, but it seems sensible to me so far. Huge amount of work @SandyRogers

workflows/ena_utils/ena_api_requests.py
SandyRogers: I would say that is slightly less precise, in that public studies are also _available_ privately (as in, with an authenticated request).

PR #42, opened 2025-02-19
added swap to megahit for SE

KateSakharova:
Added automatic change to megahit for SE

### DIFF
--- a/workflows/tests/test_assemble_study_flow.py
+++ b/workflows/tests/test_assemble_study_flow.py
@@ -46,6 +46,7 @@ def test_prefect_assemble_study_flow(
 ):
     ### ENA MOCKING ###
     accession = "SRP1"
+    number_of_runs = 3
 
     httpx_mock.add_response(
         url=f"{EMG_CONFIG.ena.portal_search_api}?"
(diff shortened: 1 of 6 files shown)


workflows/tests/test_assemble_study_flow.py
```diff
@@ -184,7 +211,14 @@ def suspend_side_effect(wait_for_input=None):
...
         analyses.models.Assembly.objects.filter(status__assembly_completed=True).count()
-        == 1
+        == 2
+    )
+    # for SE assembler should be swapped to megahit
+    assert (
+        analyses.models.Assembly.objects.filter(assembler__name='megahit').count() == 1
+    )
+    assert (
+        analyses.models.Assembly.objects.filter(assembler__name='metaspades').count() == 2
```
SandyRogers, line 221:
```suggestion
    assert (
        analyses.models.Assembly.objects.filter(assembler__name__iexact='megahit').count() == 1
    )
    assert (
        analyses.models.Assembly.objects.filter(assembler__name__iexact='metaspades').count() == 2
```

I think the mixed of the names casing may be causing trouble on this

SandyRogers (approved):
Thank you very much @KateSakharova !

I think your approach of doing this logic here is good – I wondered about having `models.Assembler.is_default_for_single_end` etc BooleanField, but this is more explicit in the flow so nice work.

PR #43, opened 2025-02-20
Submit assembly coverage-depth to ENA

SandyRogers:
Coverage-depth (not coverage) is the correct value to submit as assembly metadata, according to ENA docs.


Vkale1 (approved):
😊 ✔️

PR #44, opened 2025-02-21
Improved experiment_type

KateSakharova:
Those rules for experiment type should require library_source and library_strategy.

### DIFF
--- a/analyses/models.py
+++ b/analyses/models.py
@@ -195,7 +195,7 @@ class Run(
         latest_analysis: Analysis = self.latest_analysis
         return latest_analysis.status
 
-    def set_experiment_type_by_ena_library_strategy_and_library_source(
+    def set_experiment_type_by_metadata(
         self, ena_library_strategy: str, ena_library_source: str
     ):
         if (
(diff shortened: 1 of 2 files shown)


analyses/models.py
```diff
@@ -195,12 +195,28 @@ def latest_analysis_status(self) -> dict["Analysis.AnalysisStates", bool]:
...
         return latest_analysis.status
 
-    def set_experiment_type_by_ena_library_strategy(self, ena_library_strategy: str):
-        if ena_library_strategy.lower() == "rna-seq":
+    def set_experiment_type_by_metadata(
+        self, ena_library_strategy: str, ena_library_source: str
+    ):
+        if (
+            ena_library_strategy.lower() == "rna-seq"
+            and ena_library_source.lower() == "metagenomic"
```
SandyRogers, line 203: Only making a note here in the hopes it helps us remember later when we work on it... but it would be cool if these (semi-) controlled vocab constants like `rna-seq` etc were part of the new ENA API Handler package.

SandyRogers (approved):
Thank you Kate! 🪄

PR #45, opened 2025-02-24
Amplicon study refactor

SandyRogers:
This PR:
- refactors the amplicon analysis flow a bit, into separate files for separate tasks and subflows. Just for easier code navigation, similar to recent assembly flow changes.
- adds download-file importing for everything important in the V6 amplicon analysis results folder
- wires up the datamover for amplicon v6 analysis results, to move them over to the "transfer services area" (AKA "ftp"). Also simplified this a bit to use rsync. 
- adds an apache web server container to the dev setup, which acts as a local "transfer services area". This just serves files in the dummy /nfs/ftp area.
  - between this and the existing slurm-dev-environment psuedo datamover jobs, this means we can test out datamover + file serving jobs in a production-like way
- adds/improves more test fixtures for amplicon results, including a relatively complete set of results in the dev DB/data for a V6 amplicon analysis of MGYA00000004 
- a few other random bugs I found

All together this means the local dev environment can now serve analysis JSON and files suitable for integrating into website work.


KateSakharova (approved):
Thank you, Sandy!
I just have a small question: maybe it makes sense to create sub-folders amplicon/assembly/shared like we have in toolkit https://github.com/EBI-Metagenomics/mgnify-pipelines-toolkit/tree/main/mgnify_pipelines_toolkit/analysis ? that analyse_study_tasks repo is getting confusing..

PR #46, opened 2025-02-26
LR MIA

KateSakharova:


workflows/nextflow_utils/samplesheets.py
```diff
@@ -49,7 +49,7 @@ def queryset_to_samplesheet(
         for field in queryset.model._meta.get_fields():
             if not field.is_relation:
                 _column_map[field.name] = SamplesheetColumnSource(
-                    lookup_string=field.name
+                    lookup_string=[field.name]
```
SandyRogers, line 52: Does this need to be list-wrapped like this?

SandyRogers (approved):
Lovely stuff, thank you @KateSakharova ! One minor query inline. 

Future thought / note to self:
I suppose at some point we might want a `@property def reads_type(self) -> Literal["short", "long"]` on `Run` and `Assembly` models, so we could refactor the logic that picks `flye` over there... but until we know whether we want to save it as a DB field or not, this is perfect. I am just imagining that one day we might need it for facet search dump.

workflows/nextflow_utils/samplesheets.py
KateSakharova: but `lookup_string` is _str | list[str]_

KateSakharova: probably not, it should support string

PR #50, opened 2025-03-11
Added metat to rna-seq

KateSakharova:
RNA-Seq can be metaT as well


SandyRogers (approved):
Thank you @KateSakharova ! Is it just a common "mistake" that rna-seq can be listed as metaG?

PR #51, opened 2025-03-11
Bugfix/upload cleaned contigs

Vkale1:
Small change to upload cleaned contigs. Let me know if I missed anything. Thanks for the help!


SandyRogers (approved):
Thank you @Vkale1 !

PR #53, opened 2025-03-13
Feature/exposing results dir

MGS-sails:
This PR exposes the result_dir and metadata database columns on the analyses/{accession} endpoint

### DIFF
--- a/analyses/schemas.py
+++ b/analyses/schemas.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import logging
-from typing import List, Optional, Union
+from typing import List, Optional, Union, Dict, Any
 
 from django.conf import settings
 from ninja import Field, ModelSchema, Schema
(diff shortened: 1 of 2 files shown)


analyses/schemas.py
```diff
@@ -112,6 +112,18 @@ class MGnifyAnalysisDetail(MGnifyAnalysis):
         ],
     )
 
+    results_dir: Optional[str] = Field(
+        None,
+        description="Directory path where analysis results are stored",
+        examples=["/data/analyses/MYGA000001/results"],
```
SandyRogers, line 118:
```suggestion
        examples=["http://example.org/data/analyses/MGYA00000001/results"],
```

SandyRogers (approved):
Lovely. One tiny suggestion on the OAS example to make it a URL – otherwise merge away.

PR #54, opened 2025-03-13
Study summary generator (amplicon)

SandyRogers:
This PR:
- generates a study summary, using mgnify-pipelines-toolkit, at the end of every amplicon v6 execution for a samplesheet
- merges study summary partials across entire study at the end of all samplesheet runs
- adds `downloads` field to `Study` model and API schema, into which study summary files are pushed.
  - associated improvements to API surface for Study.


Dockerfile
SandyRogers, line 1: Only edited so we can install mgnify-pipelines-toolkit pre-release, from github.

MGS-sails (approved):
Looks good. 
Thanks @SandyRogers

PR #57, opened 2025-04-22
Clean workdirs

SandyRogers:
This PR:
- adds a flow `archive_assembly_dirs` which takes a study and will move/archive/clean its assembly dirs. Right now it just finds and delete nextflow `work` directories. In future, it may move the dirs to a backup location by adding a datamover subflow. It has some logic to not clean studies if there are unfinished assemblies.
- adds a management command to find studies suitable for cleaning. Right now that is reads studies with assemblies that haven't been updated for a while.

Future expansion:
- archiving of dirs
- use of `nextflow clean` to also clean up caches and logs if desired
- analysis studies if needed
- auto clean studies at the end of a reasonbly successful assembly flow

### DIFF
--- a/workflows/flows/assemble_study_tasks/archive_assembly_dirs.py
+++ b/workflows/flows/assemble_study_tasks/archive_assembly_dirs.py
@@ -1,0 +1,92 @@
+import operator
+from functools import reduce
+from pathlib import Path
+
+from prefect import flow, get_run_logger
+
+from activate_django_first import EMG_CONFIG
+from analyses.models import Study, Assembly
+
+
+@flow(
+    name="Archive assembly directories",
+    flow_run_name="Archive assembly directories of {reads_study_mgys}",
+)
+def archive_assembly_dirs(reads_study_mgys: str, dry_run: bool = True):
+    """
+    Currently this deletes miassembler workdirs provided the study is in a suitable state of uploaded/inactive assemblies.
+    In future, it might move assembly dirs to LTS or a results archive if needed.
+    """
+    logger = get_run_logger()
+
+    study = Study.objects.get(accession=reads_study_mgys)
+    logger.info(f"Study is {study}")
+
+    assemblies_started = study.assemblies_reads.filter_by_statuses(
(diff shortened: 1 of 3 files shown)


workflows/flows/assemble_study_tasks/archive_assembly_dirs.py
```diff
@@ -0,0 +1,92 @@
...
+                f"No directory found at {potential_workdirs_root}. Nothing to be done."
+            )
+        logger.info(f"Looking for workdirs under {potential_workdirs_root}")
+
+        if (workdir := potential_workdirs_root / "work").is_dir():
+            logger.warning(f"Found workdir {workdir}. Deleting it.")
+            if dry_run:
+                logger.info("No action since in dry_run mode.")
+            else:
+                workdir.rmdir()
```
Vkale1, line 83: I might be in old python world but does rmdir work on non empty dirs?

workflows/flows/assemble_study_tasks/archive_assembly_dirs.py
```diff
@@ -0,0 +1,92 @@
...
+    )
+
+    for potential_workdirs_root in potential_workdirs_roots:
+        if not potential_workdirs_root.is_dir():
+            logger.info(
+                f"No directory found at {potential_workdirs_root}. Nothing to be done."
+            )
+        logger.info(f"Looking for workdirs under {potential_workdirs_root}")
+
+        if (workdir := potential_workdirs_root / "work").is_dir():
```
Vkale1, line 78: Does this need to be an else? I assume if potential_workdirs_root is not a directory then it doesn't need to check folders within it

Vkale1, line 78: Does this need to be an else.. or something similar? So it doesn't search within the potential_workdirs_root dir if it's not a directory?

workflows/flows/assemble_study_tasks/archive_assembly_dirs.py
SandyRogers: 😱 You are totally right. I should have written a better test.

PR #59, opened 2025-05-06
Bugfix/merge mgys duplicates

Vkale1:
PR for merging duplicate MGYS accessions that were created as a bug by running prefect assembly and V5 analysis simultaneously. 

What it does:
All runs and assemblies are moved into the "old" V5 analysis MGYS if there is a duplicate and will delete the "new" MGYS. This ignores and throws out a warning those edge cases where a run is present in both MGYS. Those need to be fixed manually. 
We need to run the thing to pull in all "old" accessions first before running this.

Tests mock:
- a public assembly (run and assembly in different MGYS's)
- a private assembly (run and assembly in the same MGYS)
- duplicate runs in both MGYS's

Happy to understand any improvements I should make!


analyses/management/commands/merge_mgys_duplicates.py
```diff
@@ -0,0 +1,91 @@
...
+            "--dry-run",
+            action="store_true",
+        )
+
+    def handle(self, *args, **options):
+        self.dry_run = options["dry_run"]
+        self.print_duplicate_mgys_studies()
+
+    #   find ENA studies with two MGYS accessions
+    def print_duplicate_mgys_studies(self):
```
SandyRogers, line 22: Just slightly better naming since `print_...` suggests this method would be idempotent / not have side effects.
```suggestion
    def handle(self, *args, **options):
        self.dry_run = options["dry_run"]
        self.deduplicate_mgys_studies()

    #   find ENA studies with two MGYS accessions
    def deduplicate_mgys_studies(self):
```

analyses/management/commands/merge_mgys_duplicates.py
```diff
@@ -0,0 +1,91 @@
...
+            self.reassign_runs_and_assemblies(old_study, new_study)
+
+    def reassign_runs_and_assemblies(self, old_study, new_study):
+        """
+        Move runs and assemblies from new_study to old_study.
+        Delete the new_study if it's empty of runs, assemblies, and analyses.
+        """
+        # for those edge cases where both studies have the same runs, don't do anything
+        new_runs = Run.objects.filter(study=new_study)
+        old_runs = Run.objects.filter(study=old_study)
```
SandyRogers, line 51:
```suggestion
        new_runs = new_study.runs.all()
        old_runs = old_study.runs.all()
```

analyses/management/commands/merge_mgys_duplicates.py
```diff
@@ -0,0 +1,91 @@
...
+        old_run_accessions = set(a for obj in old_runs for a in obj.ena_accessions)
+        new_run_accessions = set(a for obj in new_runs for a in obj.ena_accessions)
+        if old_run_accessions & new_run_accessions:
+            logging.info(
+                f"DUPLICATE RUNS FOUND IN BOTH STUDIES: old {old_study} and new {new_study}. No further action performed."
+            )
+            return
+
+        new_assemblies = Assembly.objects.filter(assembly_study=new_study)
```
SandyRogers, line 60: This one does indeed make sense to look up this way since there are two reverse relation from assembly to study, depending on whether it is reads study or assembly study 👍

analyses/management/commands/merge_mgys_duplicates.py
```diff
@@ -0,0 +1,91 @@
...
+        Move runs and assemblies from new_study to old_study.
+        Delete the new_study if it's empty of runs, assemblies, and analyses.
+        """
+        # for those edge cases where both studies have the same runs, don't do anything
+        new_runs = Run.objects.filter(study=new_study)
+        old_runs = Run.objects.filter(study=old_study)
+        old_run_accessions = set(a for obj in old_runs for a in obj.ena_accessions)
+        new_run_accessions = set(a for obj in new_runs for a in obj.ena_accessions)
+        if old_run_accessions & new_run_accessions:
+            logging.info(
```
SandyRogers, line 55:
```suggestion
            logging.warning(
```
I think this is bad-ish?

analyses/management/commands/merge_mgys_duplicates.py
```diff
@@ -0,0 +1,91 @@
...
+
+            logging.info(
+                f"Moving {new_runs.count()} runs from {new_study} to {old_study}"
+            )
+            new_runs.update(study=old_study)
+
+        # Check if new_study is void of runs, assemblies, and analyses
+        has_runs = Run.objects.filter(study=new_study).exists()
+        has_assemblies = Assembly.objects.filter(assembly_study=new_study).exists()
+        has_analyses = Analysis.objects.filter(study=new_study).exists()
```
SandyRogers, line 82:
```suggestion
        new_study.refresh_from_db()
        has_runs = new_study.runs.exists()
        has_assemblies = Assembly.objects.filter(assembly_study=new_study).exists()
        has_analyses = new_study.analyses.exists()
```

analyses/management/commands/merge_mgys_duplicates.py
```diff
@@ -0,0 +1,91 @@
...
+
+        if not has_runs and not has_assemblies and not has_analyses:
+            if self.dry_run:
+                logging.info(f"Dry run. Real run would delete study {new_study}")
+            else:
+                new_study.delete()
+                logging.info(f"Deleted study {new_study.accession}")
+        else:
+            logging.info(f"Did not delete {new_study.accession} as study is not empty")
```
SandyRogers, line 91:
```suggestion
            logging.warning(f"Did not delete {new_study.accession} as study is not empty")
```
I think this is also bad-ish right?

analyses/tests/test_merge_mgys_duplicates.py
```diff
@@ -0,0 +1,183 @@
...
+    )
+    ena_single = ENAStudy.objects.create(
+        accession="PRJEB33333",
+        is_private=False,
+        is_suppressed=False,
+        webin_submitter="Webin-460",
+    )
+
+    # Create two MGnify studies linked to the same ENA study
+    mg_dup1_new = MGnifyStudy.objects.create(
```
SandyRogers, line 35: Are these not better called e.g. `mg_dup_a_new` ? Not a big deal obviously :)

analyses/tests/test_merge_mgys_duplicates.py
SandyRogers, line 1: Great tests, thank you!!

SandyRogers (approved):
Thank you @Vkale1 , very nice cleanup implementation! I just left a few pedantic/subjective comments as I went though.

Vkale1:
Thank you for the suggestions @SandyRogers (p.s. copilot agreed with you on all of them). All done and can be merged when you're ready.

PR #60, opened 2025-05-12
Genome and genome catalogues

MGS-sails:
This pull request involves the following:
1. The Migrations to create Genome, Genome Catalogues, and their related tables gotten from V1 APIS
2. Prefect flow to import Biomes
3. Prefect flow to import Genomes
4. APIs that expose genomes and genome-catalogues


genomes/models/GenomeCatalogue.py
```diff
@@ -0,0 +1,79 @@
...
+                  'Need not be a GOLD biome, e.g. may include host species.'
+    )
+    PROK = 'prokaryotes'
+    EUKS = 'eukaryotes'
+    VIRS = 'viruses'
+    CATALOGUE_TYPE_CHOICES = (
+        (PROK, PROK),
+        (EUKS, EUKS),
+        (VIRS, VIRS),
+    )
```
SandyRogers, line 60: This could probably now user helper `DjangoChoicesCompatibleStrEnum`

genomes/models/GenomeCatalogue.py
```diff
@@ -0,0 +1,79 @@
...
+        choices=CATALOGUE_TYPE_CHOICES,
+        max_length=20,
+    )
+    other_stats = models.JSONField(
+        db_column='OTHER_STATS_JSON', blank=True, null=True
+    )
+
+    class Meta:
+        unique_together = ('catalogue_biome_label', 'version', 'catalogue_type')
+        db_table = 'genome_catalogue'
```
SandyRogers, line 72: In general we aren't bothering to specify `db_table` in API V2 – just let django name them.

genomes/models/GenomeCatalogue.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def __str__(self):
+        return self.name
+
+    def calculate_genome_count(self):
+        self.genome_count = self.genomes.count()
+        self.save()
```
SandyRogers, line 79: See comment above re. using a GeneratedField perhaps

workflows/tests/coverage/coverage.xml
SandyRogers, line 1: this one shouldn't be checked in / needs gitignore update

workflows/tests/test_import_biomes_flow.py
SandyRogers, line 1: I think this is already covered okay by `test_importers.py`

workflows/flows/import_v5_biomes.py
SandyRogers, line 1: Cool – but any particular reason to wrap this management command in a prefect flow? I think for "one off" imports like this, a normal management command is kind of fine.

genomes/management/lib/genome_util.py
SandyRogers, line 1: Nice. I know these are mostly ported from API V1, but it would be great to refactor the file/dir sanity checkers here to use the APIV2-style validators, that is e.g. `DirectoryRule`, `GlobRule`, `FileRule` etc. I would guess copilot/cursor/junie etc would do a pretty good job of refactoring this using the helpers provided by `workflows/data_io_utils/file_rules/*`

SandyRogers (changes requested):
Thanks for this @MGS-sails ! Nice work getting to grips with the whole system.
Just a couple of suggestions to bring this PR a bit closer to the V2 patterns we've adopted:
1. Rather than all of the Downloads models, and associated base classes etc, I think we should use the `WithDownloadsModel` pattern (i.e. all of a genome's / catalogue's download objects are schematized json objects in a single jsonfield on the genome/catalogue db row). It cuts out loads of boilerplate and query JOINs.
2. Similarly with the annotations (COGs etc), I think adopting the pattern used for `annotations` field on the `Analysis` model would make a simpler data model.

I know both of these are a bit anti-relational... but I think they're reasonable choices and best to use them in Genomes as well, for consistency if nothing else.

You also mentioned needing transfer services support here – I think that can also copy the pattern used for Analysis, with a copier flow and an `external_results_dir` field on the models.

docker-compose.yaml
```diff
@@ -112,7 +112,7 @@ services:
       target: django
     container_name: app
     profiles: ["app", "all"]
-    command: "runserver 0:8000"
+    command: "runserver 0:8009"
```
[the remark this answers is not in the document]
MGS-sails: Dev debris. Will change it back

SandyRogers:
@MGS-sails one thing I forgotten in the review – a couple of test fixture files (e.g. the same ones as API V1, or some programticaly generated ones) would be useful here, to populate a couple of genomes into the dev db (i.e. included in `test_make_dev_data`). That will make it easier to work on the web client side of this feature.

emgapiv2/api/genomes.py
```diff
@@ -0,0 +1,23 @@
...
+
+router = RouterPaginated(tags=["Genomes"])
+
+
+@router.get("/", response=List[GenomeList], summary="List all genomes", operation_id="list_genomes")
+def list_genomes(request):
+    return Genome.objects.select_related("biome", "catalogue").prefetch_related("pangenome_geographic_range")
+
+
+@router.get("/{accession}", response=GenomeDetail, summary="Get genome by accession", operation_id="get_genome")
```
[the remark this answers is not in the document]
MGS-sails: This has been addressed. 
The relevant schemas have been updated and the API returned related models, similar to how it works on the Analyses app

docker-compose.yaml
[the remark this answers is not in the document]
MGS-sails: This has been removed

emgapiv2/settings.py
```diff
@@ -44,7 +44,7 @@ def show_toolbar(request):
 }
 
 
-ALLOWED_HOSTS = ["apiv2-dev.mgnify.org", "localhost", "www.ebi.ac.uk"]
+ALLOWED_HOSTS = ["apiv2-dev.mgnify.org", "localhost", "www.ebi.ac.uk", "host.docker.internal:8004"]
```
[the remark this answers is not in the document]
MGS-sails: Was used during debugging. 
Now removed

workflows/tests/coverage/coverage.xml
MGS-sails: This has been removed

workflows/tests/test_import_biomes_flow.py
MGS-sails: The test has been removed

workflows/flows/import_v5_biomes.py
MGS-sails: Not really required per say. But kept as a way to run the command in an easy when after this goes live. 
Can delete the flow afterwards

MGS-sails:
> @MGS-sails one thing I forgotten in the review – a couple of test fixture files (e.g. the same ones as API V1, or some programticaly generated ones) would be useful here, to populate a couple of genomes into the dev db (i.e. included in `test_make_dev_data`). That will make it easier to work on the web client side of this feature.



This has also been added

SandyRogers (changes requested):
Thank you @MGS-sails ! Loads of great implementation here and its a big chunk of the API V2 migration.
I left quite a lot of comments sorry, but while it looks like a lot they are mostly minor things where we can simplify, and mostly deletions! 😁 Cheers!

genomes/schemas/GenomeCatalogue.py
```diff
@@ -0,0 +1,28 @@
...
+    protein_catalogue_name: Optional[str]
+    protein_catalogue_description: Optional[str]
+    last_update: datetime
+    result_directory: Optional[str]
+    genome_count: Optional[int]
+    unclustered_genome_count: Optional[int]
+    ftp_url: str
+    pipeline_version_tag: str
+    catalogue_biome_label: str
+    catalogue_type: str
```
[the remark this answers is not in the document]
MGS-sails: Thanks @SandyRogers are you referring to only the catalogue_type variable here ?
or are you saying we should have enums for ftp_url, pipeline_version_tab and catalogue_biome_label as well ?

workflows/flows/import_genomes_flow.py
```diff
@@ -0,0 +1,338 @@
...
+
+
+@task
+def finalize_catalogue(catalogue):
+    catalogue.calculate_genome_count()
+    catalogue.save()
+
+
+@flow(name="import_genomes_flow")
+def import_genomes_flow(options: dict):
```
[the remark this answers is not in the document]
MGS-sails: Options have been exploded 💥 💥 💥

workflows/flows/import_genomes_flow.py
```diff
@@ -0,0 +1,338 @@
...
+    upload_cog_results,
+    upload_kegg_class_results,
+    upload_kegg_module_results,
+    upload_antismash_geneclusters,
+    upload_genome_files,
+)
+from genomes.models import GenomeCatalogue, Genome, GeographicLocation
+
+
+@task
```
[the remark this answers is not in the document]
MGS-sails: Done

genomes/schemas/GenomeCatalogue.py
```diff
@@ -0,0 +1,28 @@
...
+    other_stats: Optional[Dict]
+
+    class Config:
+        from_attributes = True
+
+
+class GenomeCatalogueDetail(GenomeCatalogueBase):
+    biome_id: Optional[int]
```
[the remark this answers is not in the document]
MGS-sails: A List schema has been added for GenomeCatalogue

MGS-sails:
Thanks for the quick review @SandyRogers .
I've responded to the feedback given.
Cheers !

genomes/management/lib/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+
+from django.core.management.base import BaseCommand
+from django.db.models import Q
+
+from analyses.models import Analysis
+
+logger = logging.getLogger(__name__)
+
+
+class EMGBaseCommand(BaseCommand):
```
mberacochea, line 17: This is from the old days.. I'm not sure if this is the best way of composing cli commands. I like click a lot.. so maybe - https://github.com/django-commons/django-click to make it easier to build cli commands? @SandyRogers @MGS-sails

workflows/flows/import_genomes_flow.py
```diff
@@ -0,0 +1,355 @@
...
+    return genome_data
+
+
+def finalize_catalogue(catalogue):
+    catalogue.calculate_genome_count()
+    catalogue.save()
+
+
+@flow(name="import_genomes_flow")
+def import_genomes_flow(
```
mberacochea, line 157: Are we at risk of having an invalid db connection? The import process could take some time, maybe we should refresh the db connection every now and then?

workflows/flows/import_genomes_flow.py
```diff
@@ -0,0 +1,355 @@
...
+    total = genomes.count()
+    with_annot = sum(1 for g in genomes if g.annotations)
+    with_files = sum(1 for g in genomes if g.downloads)
+    logger.info(
+        f"Final Report: {total} genomes imported. {with_annot} with annotations, {with_files} with downloads."
+    )
+    if total != with_annot:
+        logger.warning("Some genomes are missing annotations!")
+    if total != with_files:
+        logger.warning("Some genomes are missing downloads!")
```
mberacochea, line 355:
```suggestion
        logger.error("Some genomes are missing downloads!")
```

mberacochea (changes requested):
Excellent stuff @MGS-sails . I left some comments :)

workflows/flows/import_v5_biomes.py
```diff
@@ -0,0 +1,24 @@
+from prefect import flow, task
+import subprocess
+
+
+@task
+def run_import_biomes(url: str):
```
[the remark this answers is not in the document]
SandyRogers: I've a feeling we don't actually need this flow though? The idea is that:
- one-off adminy stuff (like populating the biomes to match API v1) is a boring old django management command
- regular stuff (per dataset, or on some time cadence, or event-trigger based) is a prefect flow

workflows/flows/import_v5_biomes.py
SandyRogers: Ah cool – so this management command has already been run on the Prod DB so we won't need to re-run to release this.

genomes/management/lib/__init__.py
MGS-sails: Thanks @mberacochea I checked out djclick, seems pretty nice to have. But I think we don't need to install it this time, as we're no longer using these management commands for now. Just the Prefect flow

genomes/models/genome.py
```diff
@@ -0,0 +1,199 @@
...
+    @property
+    def last_update_iso(self):
+        return self.last_update.isoformat() if self.last_update else None
+
+    @property
+    def first_created_iso(self):
+        return self.first_created.isoformat() if self.first_created else None
+
+    class Meta:
+        db_table = "genome_genomes"
```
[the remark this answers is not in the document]
MGS-sails: This is done

genomes/models/genome_catalogue.py
```diff
@@ -0,0 +1,87 @@
...
+        null=True,
+        blank=True,
+        help_text="Description of the protein catalogue, if applicable.",
+    )
+    last_update = models.DateTimeField(db_column="LAST_UPDATE", default=timezone.now)
+    result_directory = models.CharField(
+        db_column="result_directory", max_length=100, null=True, blank=True
+    )
+    biome = models.ForeignKey(
+        Biome, db_column="biome_id", on_delete=models.CASCADE, null=True, blank=True
```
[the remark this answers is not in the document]
MGS-sails: CASCADE operations have been replaced with PROTECT

workflows/flows/import_genomes_flow.py
```diff
@@ -0,0 +1,355 @@
...
+        "gold_biome": gold_biome,
+        "pipeline_version": pipeline_version,
+        "catalogue_type": catalogue_type,
+        "catalogue_biome_label": catalogue_biome_label,
+        "database": database,
+    }
+
+    options = parse_options(options)
+    catalogue = get_catalogue(options)
+    finalize_catalogue(catalogue)
```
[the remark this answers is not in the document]
MGS-sails: This function is no longer required and has been removed. As genomes count will now be a dynamic property

workflows/flows/import_genomes_flow.py
MGS-sails: @mberacochea I've added close_old_connections before genome db operations:

```python
   close_old_connections()
    genome, _ = Genome.objects.update_or_create(
        accession=accession, defaults=genome_data
    )
```

and also:

```python
    for genome_dir in genome_dirs:
        close_old_connections()
        genome_accession = process_genome_dir(catalogue, genome_dir)
        genome_accessions.append(genome_accession)
 ```

SandyRogers (changes requested):
@MGS-sails many thanks, no new comments - just flagged a few from last time that are unresolved and still relevant. None of them are big!

MGS-sails (commented):
Thanks @SandyRogers .
Please see the summary below, based on the changes we agreed on.


  🧬 Genome and Genome Catalogues Implementation

  This PR implements a comprehensive genome and genome catalogues system with the following key improvements:

  ✨ Features Added

  - Download Management: Linked download files with their indexes for better file organisation
  - API Endpoints: Complete genome and catalogue APIs with detailed responses including downloads and annotations
  - Data Models: New genome models with proper relationships, catalogues, and biome integration
  - Admin Interface: Django admin stubs for genome management

  🔧 Technical Improvements

  - Database: Replaced timestamps with base timestamped model references for consistency
  - Configuration: Consolidated variables into GenomeConfig file for better maintainability
  - Testing: Comprehensive test suite with adjusted database connection handling
  - Data Flow: Rewrote Prefect flows for genome import with JSON-based structure

  🧹 Cleanup & Refactoring

  - Deprecated old genome location models
  - Removed unused admin models and test placeholders
  - Fixed column naming issues and data format bugs

  📊 Data Management

  - Dynamic genome count on catalogue model
  - Proper primary key handling for genome_catalogues
  - Database connection refresh to prevent stale connections

SandyRogers (approved):

PR #72, opened 2025-06-10
Infer library layout from fastqs

SandyRogers:
This PR:
- updates the logic for handling incorrectly labelled library-layout metadata on ENA. We tend to prefer to work it out from the provided FASTQ files. E.g. if there is only one file, it is probably single-ended despite what the metadata says. If there are two files labelled _1 and _2, it is probably paired end even if the metadata says single.

@Ge94 the `test_get_study_readruns_from_ena` unit test is the one that will need updated.


Ge94:
I added tests for the cases covered by Sandy's change. I swapped the logic to give priority to check cases with 0 fastqs and one file named "_2" over assigning the inferred layout. 
Also, the check for no fastqs was never reached as `"".split(';')` returns `['']`, not `[]`

workflows/ena_utils/ena_api_requests.py
```diff
@@ -130,45 +130,55 @@ def get_study_from_ena(accession: str, limit: int = 10) -> ena.models.Study:
...
                 f"Incorrect names of fastq files for run {run_accession} (${sorted_fastq})"
             )
-            return None
+            return None, None
     elif len(fastq) > 2:
         logger.info(f"More than 2 fastq files provided for run {run_accession}")
-        return sorted_fastq[:2]
-    return None
+        # Try again with first two files only
+        return check_reads_fastq(sorted_fastq[:2], run_accession, library_layout)
```
KateSakharova, line 180: I'm not sure about that check. In GGP I have a different check for that case. I just check does _1 and _2 exist in those 3 files. If yes - it is PE. Otherwise, I think it should be error. I'm not sure how first 2 objects in list will look like.. It might be something similart to:
ERR_subreads.fq.gz
ERR_1.fq.gz
ERR_2.fq.gz
In that case you will check first 2 and raise error (that seems not right to me).

workflows/tests/test_ena_api_requests.py
```diff
@@ -338,6 +338,63 @@ def test_get_study_readruns_from_ena(
...
+                "host_tax_id": "7460",
+                "host_scientific_name": "Apis mellifera",
+                "instrument_platform": "ILLUMINA",
+                "instrument_model": "Illumina MiSeq",
+                "lat": "52",
+                "lon": "0",
+                "location": "hinxton",
+            },
+            {
+                "run_accession": "RUN9",
```
KateSakharova, line 380: Do we have example with 3 fqs in tests?

KateSakharova (changes requested):
Looks good! The only question is what to do with 3 files...

workflows/ena_utils/ena_api_requests.py
Ge94: The idea is that those files are sorted alphabetically first, so in your example we would have _1 and _2 first. However you might be right, I think that `.` comes before ` _ ` in ascii, which would make this fail, as we could have
ERR.fq.gz
ERR_1.fq.gz
ERR_2.fq.gz
It's worth implementing your strategy Kate IMO

workflows/ena_utils/ena_api_requests.py
```diff
@@ -130,45 +130,53 @@ def get_study_from_ena(accession: str, limit: int = 10) -> ena.models.Study:
...
-            return None
-        if "_1.f" in sorted_fastq[0] and "_2.f" in sorted_fastq[1]:
+    elif len(sorted_fastq) >= 2:
+        match_1 = next((f for f in sorted_fastq if "_1" in f), None)
+        match_2 = next((f for f in sorted_fastq if "_2" in f), None)
+        if match_1 and match_2:
             logger.info(f"Two fastqs for {run_accession}: {sorted_fastq}")
-            return sorted_fastq
+            return [match_1, match_2], "PAIRED"
         else:
```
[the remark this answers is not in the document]
Ge94: Ah I forgot about this one LOL thank you

Ge94:
@KateSakharova , help, I don't understand what I broke in the tests...

`Missing end of execution assembled runs csv file. Expected path /tmp/PRJNA1_miassembler/abc123/assembled_runs.csv.`

PR #73, opened 2025-06-11
Fix heapsizes for webin-cli assembly upload

SandyRogers:
Sets max heapsize of webin-cli jvm equal to memory available to job, and initial heapsize to half of it.


KateSakharova (approved):
what a magic 🪄

PR #75, opened 2025-06-12
Webin-cli improvements

SandyRogers:
This PR:
- adds a debug logging config for webin-cli to better understand failure cases
- adds (optional) Aspera support, where `ascp` is available. NB `ascp` must be available on PATH.


KateSakharova (approved):
You've found debug! :sandy-the-wizard:

PR #77, opened 2025-06-23
Feature/mag assembly linkage iii

MGS-sails:
1. This pull request creates the necessary DB structure to allow for linking between genomes and assemblies
2. It also exposes the /analyses/assemblies and /analyses/assemblies/<accession> endpoitns, which allows the front end to fetch mag linkage data for display
3. It also creates a prefect flow that allows linkage data to be imported from a TSV path


workflows/flows/import_genome_assembly_links.py
```diff
@@ -0,0 +1,299 @@
...
+
+    Returns:
+        Number of GenomeAssemblyLink objects created
+    """
+    logger = get_run_logger()
+    created_count = 0
+    updated_count = 0
+    skipped_count = 0
+
+    with transaction.atomic():
```
SandyRogers, line 199: nice

genomes/fixtures/genome/conftest.py
```diff
@@ -0,0 +1,171 @@
...
+        {"name": "North America"},
+        {"name": "South America"},
+        {"name": "Asia"},
+        {"name": "Africa"},
+        {"name": "Oceania"},
+        {"name": "Antarctica"},
+    ]
+    locations_objects = []
+    for location in locations_list:
+        print(f"creating geographic location {location['name']}")
```
mberacochea, line 25: Should we remove this `print` here?

genomes/fixtures/genome/conftest.py
```diff
@@ -0,0 +1,171 @@
...
+            "description": "Prokaryotic genomes from ocean environments",
+            "catalogue_biome_label": "Ocean",
+            "catalogue_type": GenomeCatalogue.PROK,
+            "biome": root_biome,
+        },
+    ]
+
+    catalogues_objects = []
+    for catalogue in catalogues_list:
+        print(f"creating genome catalogue {catalogue['name']}")
```
mberacochea, line 61: Same here

genomes/models/base_model.py
```diff
@@ -0,0 +1,10 @@
+from django.db import models
+
+
+class BaseModel(models.Model):
```
mberacochea, line 4: Don't we already have a general timestamped model?

genomes/models/genome.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+class GenomeManagerIncludingAnnotations(models.Manager):
+    def get_queryset(self):
+        return super().get_queryset()
+
+
+class Genome(WithDownloadsModel):
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
```
mberacochea, line 39: These should be in an abstract timestamped model

genomes/models/genome.py
```diff
@@ -0,0 +1,226 @@
...
+    last_update = models.DateTimeField(db_column="last_update", auto_now=True)
+    first_created = models.DateTimeField(db_column="first_created", auto_now_add=True)
+
+    geo_origin = models.ForeignKey(
+        "GeographicLocation",
+        db_column="geographic_origin",
+        null=True,
+        blank=True,
+        on_delete=models.PROTECT,
+    )
```
mberacochea, line 156: Thought: why don't we also have these in a JSON field? @MGS-sails @SandyRogers

genomes/models/genome.py
```diff
@@ -0,0 +1,226 @@
...
+    assembly = models.ForeignKey(
+        Assembly, on_delete=models.CASCADE, related_name="genome_links"
+    )
+    species_rep = models.CharField(
+        max_length=100,
+        null=True,
+        blank=True,
+        help_text="Arbitrary genome accession for species representative",
+    )
```
mberacochea, line 216: Shouldn't this one be a "Genome" too?

genomes/models/genome_catalogue.py
```diff
@@ -0,0 +1,89 @@
...
+    catalogue_type = models.CharField(
+        db_column="catalogue_type",
+        choices=CATALOGUE_TYPE_CHOICES,
+        max_length=20,
+    )
+    other_stats = models.JSONField(db_column="other_stats_json", blank=True, null=True)
+
+    class Meta:
+        unique_together = ("catalogue_biome_label", "version", "catalogue_type")
+        # db_table = "genome_catalogue"
```
mberacochea, line 82:
```suggestion
```

genomes/models/genome_catalogue.py
```diff
@@ -0,0 +1,89 @@
...
+        (PROK, PROK),
+        (EUKS, EUKS),
+        (VIRS, VIRS),
+    )
+    catalogue_type = models.CharField(
+        db_column="catalogue_type",
+        choices=CATALOGUE_TYPE_CHOICES,
+        max_length=20,
+    )
+    other_stats = models.JSONField(db_column="other_stats_json", blank=True, null=True)
```
mberacochea, line 78: Can you add a comment about what kind of "other" stats are stored here? Just to improve the readability

genomes/schemas/GenomeCatalogue.py
```diff
@@ -0,0 +1,29 @@
...
+    version: str
+    name: str
+    description: Optional[str]
+    protein_catalogue_name: Optional[str]
+    protein_catalogue_description: Optional[str]
+    last_update: datetime
+    result_directory: Optional[str]
+    genome_count: Optional[int]
+    unclustered_genome_count: Optional[int]
+    ftp_url: str
```
mberacochea, line 17: Maybe this is a dumb question... but can we use https://docs.pydantic.dev/2.1/api/networks/#pydantic.networks.AnyUrl

genomes/schemas/GenomeCatalogue.py
```diff
@@ -0,0 +1,29 @@
...
+    pipeline_version_tag: str
+    catalogue_biome_label: str
+    catalogue_type: str
```
mberacochea, line 20: Also here, we know the types

workflows/flows/import_genome_assembly_links.py
```diff
@@ -0,0 +1,303 @@
...
+
+    for i, record in enumerate(records, 1):
+        # Check for required fields
+        if not record.get("primary_assembly"):
+            logger.warning(f"Row {i}: Missing primary_assembly, skipping")
+            continue
+
+        if not record.get("genome"):
+            logger.warning(f"Row {i}: Missing genome, skipping")
+            continue
```
mberacochea, line 121: I think you can skip these two as it was already validated by `read_tsv_file`... if I'm reading this correctly

workflows/flows/import_genome_assembly_links.py
```diff
@@ -0,0 +1,303 @@
...
+
+        # Clean up the record
+        cleaned_record = {
+            "primary_assembly": record["primary_assembly"].strip(),
+            "genome": record["genome"].strip(),
+            "mag_accession": (
+                record.get("mag_accession", "").strip()
+                if record.get("mag_accession")
```
mberacochea, line 129: Pedantic comment:
```
record["species_rep"].strip()
if "species_rep" in record
else None
```

workflows/flows/import_genome_assembly_links.py
```diff
@@ -0,0 +1,303 @@
...
+
+        # Find the Assembly
+        try:
+            assembly = Assembly.objects.get_by_accession(record["primary_assembly"])
+        except (ObjectDoesNotExist, Assembly.MultipleObjectsReturned):
+            logger.warning(
+                f"Assembly not found or multiple found: {record['primary_assembly']}"
+            )
+
+        results.append((record, genome, assembly))
```
mberacochea, line 179: This follows Sandy comment about performance, this is likely to make loads of queries. I think we need to batch these with `IN` instead.

mberacochea (changes requested):
Thanks @MGS-sails I left a few comments.

I agree with Sandy, we should test this with a large TSV (I feel we will bump into some performance issues with the current code)

genomes/models/genome.py
MGS-sails: @mberacochea we do. 
But this branch was created prior to that work being done. 
But I've made that change on my genomes api branch which has already been merged

genomes/models/genome.py
MGS-sails: @mberacochea  this has also been implemented in a different PR

genomes/models/genome_catalogue.py
MGS-sails: This has also been done on the genome catalogue api PR

genomes/models/genome.py
MGS-sails: Not clear on what you mean here. 
But species_rep was added as it's needed for MAG assembly linkage

emgapiv2/test_api.py
```diff
@@ -371,3 +371,95 @@ def test_biomes_endpoint(ninja_api_client, top_level_biomes):
...
+def test_api_assembly_genome_links_empty(mgnify_assemblies, ninja_api_client):
+    # Ensure an accession is set for the first assembly
+    assembly = mgnify_assemblies[0]
+    assembly.ena_accessions = ["ERZ1001"]
+    assembly.save()
+
+    items = call_endpoint_and_get_data(
+        ninja_api_client,
+        f"/assemblies/{assembly.ena_accessions[0]}/genome-links",
+        count=0,
```
mberacochea: Wouldn't be better to have test data with not links instead of using `count=0`?

emgapiv2/test_api.py
```diff
@@ -371,3 +371,95 @@ def test_biomes_endpoint(ninja_api_client, top_level_biomes):
...
+    )
+    assert isinstance(items, list)
+    assert items == []
+
+
+@pytest.mark.django_db
+def test_api_assembly_genome_links_with_data(
+    mgnify_assemblies, genomes, ninja_api_client
+):
+    from genomes.models import GenomeAssemblyLink
```
mberacochea: Move the import to the top please

genomes/models/base_model.py
```diff
@@ -0,0 +1,11 @@
...
+    id = models.AutoField(primary_key=True)
+    # created_at = models.DateTimeField(auto_now_add=True)
```
mberacochea:
```suggestion
```

mberacochea: This exists already in the analysis app:
```
class TimeStampedModel(models.Model):
    created_at = models.DateTimeField(auto_now_add=True)
    updated_at = models.DateTimeField(auto_now=True)

    class Meta:
        abstract = True
```

workflows/flows/import_genome_assembly_links_flow.py
```diff
@@ -0,0 +1,385 @@
...
+
+    Args:
+        tsv_path: Path to the TSV file
+
+    Returns:
+        The validated path to the TSV file
+
+    Raises:
+        FileNotFoundError: If the TSV file does not exist
+        ValueError: If the path is not a file
```
mberacochea: We use https://peps.python.org/pep-0287/ for docstrings (not a blocker)

workflows/flows/import_genome_assembly_links_flow.py
```diff
@@ -0,0 +1,385 @@
...
+    assembly_accessions = {
+        r["primary_assembly"] for r in records if r.get("primary_assembly")
+    }
+
+    genomes_map = {
+        g.accession: g
+        for g in Genome.objects.filter(accession__in=list(genome_accessions)).only(
+            "genome_id", "accession"
+        )
+    }
```
mberacochea: Maybe it's a good idea to count the number of genomes to make sure we ar getting the expected total?

workflows/flows/import_genome_assembly_links_flow.py
```diff
@@ -0,0 +1,385 @@
...
+
+    # Assembly has custom manager method get_by_accession for single, but here we batch by ena_accessions or similar.
+    # Prefer a generic filter that supports accession lookup via a field used by get_by_accession.
+    # Fallback: use get_by_accession per accession if bulk filter is not straightforward.
```
mberacochea: I think this logic belongs in the model manager.

genomes/models/base_model.py
MGS-sails: I've removed the redundant genome base model. Resorted to using the Analysis timestamped model

genomes/management/commands/make_model.py
```diff
@@ -0,0 +1,95 @@
+from django.core.management.base import BaseCommand
```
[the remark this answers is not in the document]
MGS-sails: Happy to create the app for this going forward. 
However for now, I thought the management/commands is good temp home for them ?
Let me know if you have a better location in mind

workflows/flows/import_genome_assembly_links_flow.py
```diff
@@ -0,0 +1,357 @@
...
+        return results
+
+    genome_accessions = {r["genome"] for r in records if r.get("genome")}
+    assembly_accessions = {
+        r["primary_assembly"] for r in records if r.get("primary_assembly")
+    }
+
+    genomes_map = {
+        g.accession: g
+        for g in Genome.objects.filter(accession__in=list(genome_accessions)).only(
```
mberacochea: Just for reference, django has a method called `.value_list` does something similar.

workflows/flows/import_genome_assembly_links_flow.py
```diff
@@ -0,0 +1,357 @@
...
+
+@task(name="Find Genome and Assembly objects")
+def find_objects(
+    records: List[Dict[str, str]],
+) -> List[Tuple[Dict[str, str], Optional[Genome], Optional[Assembly]]]:
+    """
+    Find Genome and Assembly objects for each input record using batched lookups
+    to reduce database round trips.
+
+    :param records: Validated and cleaned records.
```
mberacochea: Not important: there is mixture of docstrings types.. we need to normalize them

mberacochea (changes requested):
I left some comments, none super important I think.

My only concern is that some of these flows are very DB intensive and the input TSV may be largue. In that case it is worth it to add some `close_old_connections()` after long processing to avoid stale db connections. Django outside of the request cycle doesn't do a good job and automatically testing this is a bit tricky.

emgapiv2/test_api.py
```diff
@@ -371,3 +371,95 @@ def test_biomes_endpoint(ninja_api_client, top_level_biomes):
...
     call_endpoint_and_get_data(ninja_api_client, "/biomes/?max_depth=2", count=3)
     call_endpoint_and_get_data(ninja_api_client, "/biomes/?max_depth=10", count=4)
+
+
+@pytest.mark.django_db
+def test_api_assemblies_list(mgnify_assemblies, ninja_api_client):
+    # Set accessions for the assemblies
+    for i, assembly in enumerate(mgnify_assemblies):
+        assembly.ena_accessions = [f"ERZ{i+1}"]
+        assembly.save()
```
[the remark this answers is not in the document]
MGS-sails: @mberacochea i've added a fixture for this

MGS-sails:
@mberacochea please see my responses to https://github.com/EBI-Metagenomics/emgapi-v2/pull/77#discussion_r2572294319

PR #78, opened 2025-06-25
Add support for miassembler v3.0.1

mberacochea:
I've added also the "contaminant_reference" in the assemblies metadata as this could be useful to query later.

All assemblies should be decontaminated of human data, that is done in the pipeline configuration. But't it's possible to specify a different human reference of phix reference (not likely but... you never know)

### DIFF
--- a/workflows/tests/test_assemble_study_flow.py
+++ b/workflows/tests/test_assemble_study_flow.py
@@ -16,7 +16,7 @@ import ena.models
 from workflows.flows.assemble_study import AssemblerChoices, assemble_study
 from workflows.flows.assemble_study_tasks.assemble_samplesheets import (
     get_reference_genome,
-    update_assemblies_and_contaminant_ref_assemblers_from_samplesheet,
+    update_assemblers_and_contaminant_ref_of_assemblies_from_samplesheet,
 )
 from workflows.flows.assemble_study_tasks.make_samplesheets import (
     make_samplesheets_for_runs_to_assemble,
@@ -617,7 +617,7 @@ def test_assembler_changed_in_samplesheet(
     # run assembly on samplesheet
     # note that assembler arg will be metaspades
     ss_df = pd.read_csv(samplesheet)
-    update_assemblies_and_contaminant_ref_assemblers_from_samplesheet(ss_df)
+    update_assemblers_and_contaminant_ref_of_assemblies_from_samplesheet(ss_df)
 
     # flow should have updated assemblies to have megahit assembler, as per edited samplesheet
     assert (


analyses/models.py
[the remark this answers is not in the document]
mberacochea: Maybe.. in the django world people tend to have a few massive model files.

workflows/flows/assemble_study_tasks/assemble_samplesheets.py
```diff
@@ -165,19 +164,33 @@ def update_assemblies_assemblers_from_samplesheet(
...
+            contaminant_reference = assembly_row.get("contaminant_reference")
+            if contaminant_reference and not pd.isna(contaminant_reference):
+                latest_assembly.metadata[
+                    analyses.models.Assembly.CommonMetadataKeys.CONTAMINANT_REFERENCE
+                ] = contaminant_reference
+                update = True
+        except KeyError:
+            logger.warning("Could not find contaminant reference in samplesheet")
+
+        if update:
```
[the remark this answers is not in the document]
mberacochea: I was trying to save one query, if there is no `contaminant_reference` then there is no need for the update.

workflows/flows/assemble_study_tasks/assemble_samplesheets.py
```diff
@@ -136,19 +135,19 @@ def update_assembly_metadata(
 
 
 @task
-def update_assemblies_assemblers_from_samplesheet(
+def update_assemblies_and_contaminant_ref_assemblers_from_samplesheet(
```
SandyRogers, line 138: Kinda confusing name now, maybe:

```suggestion
def update_assemblers_and_contaminant_ref_of_assemblies_from_samplesheet(
```

workflows/flows/assemble_study_tasks/assemble_samplesheets.py
```diff
@@ -165,19 +164,33 @@ def update_assemblies_assemblers_from_samplesheet(
                 f"Could not find unique assembly for {assembly_row['reads_accession']}"
             )
             continue
```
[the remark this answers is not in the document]
SandyRogers: Good point - probably we had a .get() previously and switched to .filter().first() later

workflows/flows/assemble_study_tasks/assemble_samplesheets.py
```diff
@@ -136,19 +135,19 @@ def update_assembly_metadata(
...
+    Updates the assemblers and contaminant reference associated with each assembly, based on what is in the samplesheet.
     This is done because a samplesheet CSV may have been edited since first created.
     :param samplesheet_df: Pandas dataframe of the samplesheet content
     :return:
     """
     logger = get_run_logger()
     for _, assembly_row in samplesheet_df.iterrows():
         try:
-            latest_assembly = (
+            latest_assembly: analyses.models.Assembly = (
```
SandyRogers, line 150:
```suggestion
            latest_assembly: analyses.models.Assembly | None = (
```

workflows/flows/assemble_study_tasks/assemble_samplesheets.py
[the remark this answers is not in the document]
SandyRogers: Doesn't need to be now, but we should consider adopting [django-dirtyfields](https://django-dirtyfields.readthedocs.io/en/stable/) so we can just do `if latest_assembly.is_dirty(): latest_assembly.save()`.
Obviously not worth a library for this one use, but we have many defensive / potentially unneeded `.save()`s scattered around the `workflows/` app.

workflows/flows/assemble_study_tasks/make_samplesheets.py
```diff
@@ -114,6 +130,20 @@ def make_samplesheet(
...
+            # sets these for the whole samplesheet. Also, if the human_reference global parameter and human_reference
+            # in the samplesheet are empty, the pipeline will fail (the user needs to provide a
+            # skip_human_decontamination flag).
+            # Reference doc -> https://github.com/EBI-Metagenomics/miassembler?tab=readme-ov-file#usage
+            "human_reference": SamplesheetColumnSource(
+                lookup_string="id", renderer=lambda _: ""
+            ),
+            "phix_reference": SamplesheetColumnSource(
+                lookup_string="id", renderer=lambda _: ""
+            ),
```
SandyRogers, line 146: We should probably change `SamplesheetColumnSource` to allow const rendering like

```python
class SamplesheetColumnSource(BaseModel):
    lookup_string: str | list[str] = Field(default="id")
    renderer: Callable[[Any], str] = Field(lambda x: x)
    const: Any | None = Field(None)
```

because it is a bit more readable to do

```python
            "assembly_memory": SamplesheetColumnSource(
                const=memory
            ),
            "human_reference": SamplesheetColumnSource(
                const=""
            ),
```

-> Does not need to be in this pr.

mberacochea: haha I wrote one of those and the roll it back

workflows/flows/assemble_study_tasks/assemble_samplesheets.py
mberacochea: Java, is that you?

MGS-sails (approved):

PR #88, opened 2025-07-01
Add WGA as an allowed library strategy for METAGENOMIC and METATRANSCRIPTOMIC experiment types

chrisAta:
We used to accept `WGA` as a valid library strategy, this PR brings that back:
- If `WGA` + `library_source=metagenomic`, then it's a `METAGENOMIC` experiment type
- If `WGA` + `library_source=metatranscriptomic`, then it's a `METATRANSCRIPTOMIC` experiment type


Vkale1 (approved):
Cool! :)

KateSakharova (approved):
Looks good!
Ideally, you was right, we should have test for `experiment_type` model. I will not force you doing that :)

PR #89, opened 2025-07-02
Add fallback to ena data portal if metagenome data portal returns no data

ochkalova:
- ENAAPIRequest.get() method updated to make multiple requests until first non-empty response
- default portals are now set in config 
- new test

### DIFF
--- a/workflows/tests/test_ena_api_requests.py
+++ b/workflows/tests/test_ena_api_requests.py
@@ -124,6 +124,34 @@ def test_get_study_from_ena_use_secondary_as_primary(httpx_mock, prefect_harness
 
 @pytest.mark.httpx_mock(should_mock=should_not_mock_httpx_requests_to_prefect_server)
 @pytest.mark.django_db(transaction=True)
+def test_get_study_only_available_in_ena_portal(httpx_mock, prefect_harness):
+    """
+    Required study is in ena portal, but not in metagenome portal
+    """
+    sec_study_accession = "SRP0009034"
+    httpx_mock.add_response(
+        url=f"{EMG_CONFIG.ena.portal_search_api}?result=study&query=%22%28study_accession={sec_study_accession}%20OR%20secondary_study_accession={sec_study_accession}%29%22&limit=10&format=json&fields={','.join(EMG_CONFIG.ena.study_metadata_fields)}&dataPortal=metagenome",
+        json=[],
+    )
+    httpx_mock.add_response(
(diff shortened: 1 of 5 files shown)


workflows/tests/test_assemble_study_flow.py
```diff
@@ -396,7 +396,7 @@ def suspend_side_effect(wait_for_input=None):
         analyses.models.Assembly.objects.filter(
             assembler__name__iexact="metaspades"
         ).count()
-        == 2
+        == 4
```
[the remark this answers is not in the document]
SandyRogers: Thank you kindly! I also configured an extra branch protection rule as the other PR shouldn't have been able to merge to `main` with a failing test.

workflows/tests/test_ena_api_requests.py
```diff
@@ -122,6 +122,39 @@ def test_get_study_from_ena_use_secondary_as_primary(httpx_mock, prefect_harness
     assert len(created_study.additional_accessions) == 1
 
 
+@pytest.mark.httpx_mock(should_mock=should_not_mock_httpx_requests_to_prefect_server)
```
SandyRogers, line 125: I *think* in this case you won't need the `should_mock` filter, as no prefect server test harness fixture is running here. But it is also fine to leave it as it does no harm :) 

```suggestion
@pytest.mark.httpx_mock
```

workflows/tests/test_ena_api_requests.py
```diff
@@ -122,6 +122,39 @@ def test_get_study_from_ena_use_secondary_as_primary(httpx_mock, prefect_harness
...
+    sec_study_accession = "SRP0009034"
+    httpx_mock.add_response(
+        url=f"{EMG_CONFIG.ena.portal_search_api}?result=study&query=%22secondary_study_accession={sec_study_accession}%22&limit=10&format=json&fields={','.join(EMG_CONFIG.ena.study_metadata_fields)}&dataPortal=metagenome",
+        json=[],
+    )
+    httpx_mock.add_response(
+        url=f"{EMG_CONFIG.ena.portal_search_api}?result=study&query=%22secondary_study_accession={sec_study_accession}%22&limit=10&format=json&fields={','.join(EMG_CONFIG.ena.study_metadata_fields)}&dataPortal=ena",
+        json=[
+            {
+                "study_title": "I wanted to be normal, but they put me in ena portal",
```
SandyRogers, line 140: 🥹

workflows/tests/test_ena_api_requests.py
ochkalova: Yes, you're right. I just copied it without thinking...

PR #90, opened 2025-07-02
fix for assembler selection

KateSakharova:
Still using wrong library_layout in assembly flow :facepalm:


workflows/flows/assemble_study_tasks/make_samplesheets.py
```diff
@@ -107,29 +107,42 @@ def make_samplesheet(
...
                     else (
                         analyses.models.Assembler.SPADES
-                        if layout == SINGLE_END_LIBRARY_LAYOUT
-                        and platform
+                        if metadata.get(
+                            analyses.models.Run.CommonMetadataKeys.INFERRED_LIBRARY_LAYOUT,
+                            metadata.get(
+                                analyses.models.Run.CommonMetadataKeys.LIBRARY_LAYOUT
+                            ),
+                        )
```
SandyRogers, line 131: Not for this PR... but in case you or I end up adding more such `INFERRED_*` metadata fields, I think we should add an
```python
@property
def inferred_metadata()
```
to the model. Then there we can just overwrite the value of `LIBRARY_LAYOUT` with `INFERRED_LIBRARY_LAYOUT` if it is present. So places like this can just do `run.inferred_metadata.get(LIBRARY_LAYOUT)` and get the value, whether or not it was inferred. All these places should not have to care about where the value came from I think

workflows/tests/test_assemble_study_flow.py
```diff
@@ -244,6 +256,44 @@ def test_prefect_assemble_study_flow(
...
+                "host_scientific_name": "Apis mellifera",
+                "instrument_platform": "ILLUMINA",
+                "instrument_model": "Illumina MiSeq",
+                "lat": "52",
+                "lon": "0",
+                "location": "hinxton",
+            },
+            {
+                "sample_accession": "SAMN09",
+                "sample_title": "Wookie hair 9 (SE labeled as PE, should use changed LL)",
```
SandyRogers, line 280: so many wookies

workflows/tests/test_assemble_study_flow.py
```diff
@@ -272,67 +322,52 @@ def suspend_side_effect(wait_for_input=None):
...
-    )
-    os.makedirs(
         f"{assembly_folder}/PRJNA1/PRJNA1/SRR7/SRR7/assembly/metaspades/3.15.5/coverage/",
-        exist_ok=True,
-    )
+        f"{assembly_folder}/PRJNA1/PRJNA1/SRR8/SRR8/assembly/metaspades/3.15.5/coverage/",
+        f"{assembly_folder}/PRJNA1/PRJNA1/SRR9/SRR9/assembly/megahit/1.2.9/coverage/",
+    ]
+
+    for folder in created_coverage_folders:
```
SandyRogers, line 345: lovely

SandyRogers (approved):
Thank you @KateSakharova !!

PR #91, opened 2025-07-04
Adds "policy" for library_strategy metadata when assembling 

SandyRogers:
- Can add optional "policy" to assembly flow, commonly used to include library-strategy=OTHER data in the samplesheets for assembly
- Refactors the queryset -> samplesheet utils a bit, adding a simpler way to stick constants into a samplesheet (like a single memory value on all rows) as well as adding metadata that depends on a model instance's python methods (e.g. a `@property`) instead of a db lookup string
- adds an `inferred_metadata` property on various models, via a mixin, which is just the `metadata` jsonfield/dict, but if a metadata key for `my_thing` happens to exist with name `inferred_my_thing`, that inferred value is used instead. This is so we can keep track of original/ENA metadata values, as well as our own inferred ones e.g. incorrectly asserted SINGLE/PAIRED layouts, but be sure we are using the inferred ones without complicated lookups.

### DIFF
--- a/.talismanrc
+++ b/.talismanrc
@@ -149,3 +149,4 @@ allowed_patterns:
 - KnownMetadataKeys
 - primary_key
 - pass_whole_object
+- dictionary key
--- a/README.md
+++ b/README.md
@@ -163,8 +163,62 @@ This is bidi: it creates a `manage.py prefectcli` command to run Prefect, and it
 
 The API is implemented with `ninja` (`emgapiv2/api.py`), and uses Open API spec with Swagger.
 
+#### Model Helpers and Utilities
+
+The `analyses/base_models/base_models.py` file contains several useful mixins and base classes:
+
+- `GetByENAAccessionManagerMixin`: Since many MGnify models share or relate to ENA objects with multiple accessions (e.g. primary/secondary), this provides a method to get objects by any of their ENA accessions.
+- `UpdateOrCreateByAccessionManagerMixin`: A more useful way to get/create objects by accession whilst merging/setting metadata.
+- `InferredMetadataMixin`: A mixin for models with a `metadata={}` field that may contain an original dictionary key (e.g. `library_strategy`) which we know has incorrect metadata, so we also include an `inferred_library_strategy` dictionary key. This mixin provides a `metadata_preferring_inferred` property.
+- `ENADerivedModel`: A base model for objects derived from ENA objects – especially those which should follow ENA accessions/visibility/privacy/suppression etc. Look for Django signals based on this.
+
 ### Writing flows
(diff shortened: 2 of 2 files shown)


workflows/flows/assemble_study_tasks/make_samplesheets.py
```diff
@@ -123,48 +123,34 @@ def make_samplesheet(
...
-                            analyses.models.Run.CommonMetadataKeys.INFERRED_LIBRARY_LAYOUT,
-                            metadata.get(
-                                analyses.models.Run.CommonMetadataKeys.LIBRARY_LAYOUT
-                            ),
+                        if assembly.run.inferred_metadata.get(
+                            analyses.models.Run.CommonMetadataKeys.LIBRARY_LAYOUT
                         )
                         == SINGLE_END_LIBRARY_LAYOUT
-                        and metadata.get(
+                        and assembly.run.inferred_metadata.get(
```
KateSakharova, line 130: Maybe I understood not whole code correctly.. but my first question was: how I should know does field have `inferred` record or not? We know in advance that we have implemented inferred for `library_layout` but if you will add it to `library_strategy`, for example, how I should find/check it? 
In my understanding inferred function should work in a bit different way: I should use original model name and then model fuction should decide: use inferred value (if that exists) or original. So I will not need to check does `inferred` exists of not. 
I'm not sure it does make sense to implement in Django terms.. but maybe something to think about

KateSakharova (approved):
Nice, Sandy! I think I will need to poke around prefect to understand fully all those policy additions. I have left one question.

workflows/flows/assemble_study_tasks/make_samplesheets.py
SandyRogers: Thanks @KateSakharova . This does indeed work how you wish it to, in that `inferred_metadata` is a sort of clever version of `metadata`, but when a key within `metadata` starts with `inferred_` then that it preferred to one that does not.
A bit hard to follow in the implementation code I think, but in the tests it is clearer: https://github.com/EBI-Metagenomics/emgapi-v2/pull/91/files#diff-364fcc141694a145044a135f232afceaa22c852cb1f62a242b6ee4e7235213d7R460

I considered overriding the `.metadata` itself, but that seemed a bit dangerously magical and likely to break something in django internals potentially.

But yes, as it stands you can just do `model.inferred_metadata["something"]` and not care about whether we have `inferred_something` or not.

SandyRogers: Good point :) I updated README with details of a bunch of the helper things we have (ones that I could remember anyway)

PR #94, opened 2025-07-08
Gracefully handle completely failed amplicon sheets

SandyRogers:
If **no** amplicon analysis runs complete successfully, do not try (and therefore fail) to run summary generator commands. Instead, exit happily.


Ge94 (approved):
Looks straightforward :)

PR #97, opened 2025-07-11
workdir should be workDir

KateSakharova:


SandyRogers:
@KateSakharova The plot thickens:
```
➜  nextflow run -help | grep work
Nextflow 25.04.6 is available - Please consider updating your version to it
       Entry workflow name to be executed
       Directory where workflow outputs are stored
       Run the workflow script skipping the execution of all processes
       Execute the workflow replacing process scripts with command stubs
       Send a notification email on workflow completion to the specified
       Monitor workflow execution with Seqera Platform (formerly Tower Cloud)
       Send workflow status messages via HTTP to target URL
    -w, -work-dir
    ```

SandyRogers (approved):

PR #102, opened 2025-07-16
Use rawreads codon nextflow config defined in emgapi-v2 config.py

timrozday-mgnify:
Raw reads pipeline needs to use a codon nextflow config located in pipelines-nextflow-configs (it is no longer in the raw-reads-analysis-pipeline repo). 

I have 
* added a line in the emgapi-v2 config pointing to our local (hps) clone of the pipelines-nextflow-configs repo
* changed the flow to use that config

### DIFF
--- a/emgapiv2/config.py
+++ b/emgapiv2/config.py
@@ -102,7 +102,7 @@ class RawReadsPipelineConfig(BaseModel):
     rawreads_pipeline_git_revision: str = (
         "master"  # branch or commit of ebi-metagenomics/raw-reads-analysis-pipeline
     )
-    rawreads_pipeline_config_file: str = "/nfs/production/nextflow-configs/codon.conf"
+    rawreads_pipeline_config_file: str = "/nfs/production/nextflow-configs/codon.config"
     rawreads_pipeline_nf_profile: str = "codon"
     samplesheet_chunk_size: int = 50
     # results stats


emgapiv2/config.py
```diff
@@ -102,6 +102,9 @@ class RawReadsPipelineConfig(BaseModel):
     rawreads_pipeline_git_revision: str = (
         "master"  # branch or commit of ebi-metagenomics/raw-reads-analysis-pipeline
     )
+    rawreads_pipeline_config_file: str = (
+        "/hps/software/users/rdf/metagenomics/service-team/configs/pipelines-nextflow-configs/pipelines/raw-reads-analysis-pipeline/codon.conf"
```
KateSakharova, line 106: I think that path should be "fake" that works for tests. Real production path should be on cluster config. I will point you to it. 
And additional note: could you rename file `config` instead of `conf` for consistency in configs repo  :)

KateSakharova (changes requested):

emgapiv2/config.py
timrozday-mgnify: Ah OK, sandy jsut explained how the secrets file works. I'll change this to a dummy path

I made a mistake with the extension (`.conf` vs `.config`). It is `.config` in pipelines-nextflow-configs

KateSakharova (approved):
Brilliant! Thank you!

PR #106, opened 2025-07-22
print assembled_runs_csv into logging

KateSakharova:
Print list of assembled runs into logging. That will help resume mechanism. No need to modify samplesheet for those runs if they are done.


SandyRogers (approved):
Cool! Thank you @KateSakharova :)

PR #121, opened 2025-08-26
Analyses get_or_create should not consider privacy in query

SandyRogers:
Previously, if a study/run was created, and then its privacy/owner was changed (e.g. due to a mistake in initial submission), than duplicate analyses were created in the v6 analysis flows, since the get_or_create included the privacy/ownership in the query. This PR makes these "defaults", i.e. not queries for but set on first creation. Other hooks etc should be responsible for propagating privacy/ownership changes between models.


KateSakharova (approved):
Thank you, Sandy!

PR #136, opened 2025-10-01
Prefect slurm parallel workers

arajkovic:
This PR is a modification of the way flows are run based on the type of worker. The idea behind this is to be able to run both the old process type of worker and the new prefect-slurm worker in parallel.
Once we are confident enough that the prefect-slurm worker is stable and working fine, flows should be modified to take into account the new way of running things.

Note on the environment patching:
After updating slurm and PySlurm to version 24, passing `environment="ALL"` as a kwarg does not seem to work anymore. To work around this, I implemented a helper function to mimic the behaviour of the `--export` flag in `sbatch`. Again, once we have just the prefect-slurm worker running, all of this will become redundant since the environment will be defined via the deployment and not inherited from the calling process.


workflows/prefect_utils/slurm_flow.py
```diff
@@ -300,6 +306,201 @@ def start_or_attach_cluster_job(
...
+    )
+
+    try:
+        _stdout, stderr = process.communicate(timeout=expected_time.total_seconds())
+    except TimeoutError as e:
+        logger.error(f"Failed to run subprocess: {e}")
+        process.kill()
+        process.wait()
+        ocj.last_known_state = SlurmStatus.failed
+        ocj.updated_at = datetime.now()
```
SandyRogers, line 477:
```suggestion
```
it is an `auto_now` field already

workflows/prefect_utils/slurm_flow.py
```diff
@@ -300,6 +306,201 @@ def start_or_attach_cluster_job(
...
+        ocj.save()
+
+        raise SubprocessFailedException(process.pid, process.returncode, "Timed out")
+
+    if process.returncode == 0:
+        logger.info(f"Job {ocj} finished successfully.")
+
+        ocj.last_known_state = SlurmStatus.completed
+        ocj.updated_at = datetime.now()
```
SandyRogers, line 486:
```suggestion
```

workflows/prefect_utils/slurm_flow.py
```diff
@@ -300,6 +306,201 @@ def start_or_attach_cluster_job(
...
+        ocj.save()
+
+        store_nextflow_trace(ocj)
+    else:
+        ocj.last_known_state = SlurmStatus.failed
+        ocj.updated_at = datetime.now()
```
SandyRogers, line 492:
```suggestion
```

workflows/prefect_utils/slurm_flow.py
```diff
@@ -300,6 +306,201 @@ def start_or_attach_cluster_job(
...
+    This task MAY launch a subprocess, otherwise it may return the Job ID of a previously launched job
+    that is considered identical.
+
+    This allows flows to "reattach" to slurm jobs that they previously started,
+    even if the flow has crashed and been restarted.
+    (E.g. if the prefect worker VM is restarted during a long-running nextflow pipeline.)
+
+    It also allows flows to require a slurm job to have run, but to accept that slurm job may have been run by
+    a previous or different flow.
+    (E.g. if a metagenome assembly is needed by two different analysis pipelines.)
```
SandyRogers, line 335:
```suggestion
    This task MAY launch a subprocess, otherwise it may return a database object of a previously run job that is considered identical (e.g. the same command was run for the same inputs).

    This allows flows to "reattach" to command that they previously ran,
    even if the flow has crashed and been restarted. In this case, it could be that the command previously finished successfully, or failed, and is not worth retrying – so by reattach we mean use the previous result. It could also mean that the exact same command is currently running under a different flow, in which case this method MAY currently fail since the `pid` of that currently running job may well be unavailable to this flowrun (if it is on a different slurm node).

    It also allows flows to require a slurm job to have run, but to accept that slurm job may have been run by
    a previous or different flow.
    (E.g. if a metagenome assembly is needed by two different analysis pipelines.)
```

SandyRogers: I believe this is a more accurate representation of the current design + limitations?

SandyRogers (commented):
Absolutely great, thank you @arajkovic ! I left a few suggestions/queries throughout. Also, did you make an alternative version of the "Realistic example" flow run that uses the prefect-worker _explicitly_, or is that future work out of scope for this PR?

workflows/prefect_utils/slurm_flow.py
```diff
@@ -421,6 +636,22 @@ def run_cluster_job(
     for unsafe in SLURM_UNSAFE_CHARS:
         _name = _name.replace(unsafe, "-")
 
+    worker_type = get_prefect_worker_type()
+
+    if worker_type == "slurm":
```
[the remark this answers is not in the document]
arajkovic: yeah, makes more sense

Dockerfile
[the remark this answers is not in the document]
arajkovic: 🙇

docker-compose.yaml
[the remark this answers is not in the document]
arajkovic: Thanks

README.md
```diff
@@ -132,6 +132,8 @@ task prefect -- deployment run "Download a study read-runs/realistic_example_dep
 
 (Note that you can't run this one in the same way as `simple_example.py`, because `realistic_example.py` does not have a `__main__`).
 
+Note: If you get an error like `I/O error writing script/environment to file` while running the flow, make sure to add write permissions for other users to the working directory of the flow `chmod -R o+x slurm-dev-environment/fs/nfs/production/dev-slurm-work-dir`
```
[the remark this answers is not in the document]
arajkovic: Yeah, makes sense. Should I do it recursively for the whole filesystem under `slurm-dev-environment/fs`?
@SandyRogers

arajkovic:
@SandyRogers
> Absolutely great, thank you @arajkovic ! I left a few suggestions/queries throughout. Also, did you make an alternative version of the "Realistic example" flow run that uses the prefect-worker _explicitly_, or is that future work out of scope for this PR?


Thanks!
I added an alternative example showing parallel execution of tasks.

SandyRogers (approved):
Great additions there @arajkovic thank you. :shipit:

PR #137, opened 2025-10-02
Override mpt version default in amplicon pipeline

Ge94:


mberacochea:
Why do we need this one? IMO this makes things a bit messy, many places for the same setting.

It should be toolkit update -> amplicon update -> prefect picks up this (automatically would be ideal)

Ge94:
@mberacochea I agree this is mostly a workaround because of the way the amplicon pipeline is structured. We'll probably need to work more on that side, I am okay to leave prefect as is

PR #199, opened 2025-12-03
This is a test I want to run - as_subflow=False.

mberacochea:
I'm adding this because timeout=0 didn't do the trick and the assembly_study flow is getting "stuck" after running the batches.


SandyRogers (commented):
According to the prefect docs, this `as_subflow` should NOT affect the synchronicity of the return. It should just affect whether or not the subflow is linked in a metadata/flow hierarchy sense. If the `timeout=0` is not working, I wonder if this is because the call blocks until a flow-run ID can be returned.. perhaps something is preventing that? It should not be that the subflow needs to start (which would obviously also block until a slurm job can be created for the subflow deployment run). So I'm not quite sure... is there definitely an active deployment created for this subflow?

In general this looks like a good candidate for adding something to the hello-nextflow test flow to more rapidly debug on the local donco cluster.

PR #250, opened 2026-01-26
Assembly analysis V6 out-of-production ingestion flow

ochkalova:
### Overview

**Commit history is messy, because I worked on this for a long time and some features were added and then reverted. Please don't rely on commits for guidance.** 

This PR adds `import_out_of_production_assembly_analysis_results`, a new Prefect flow that ingests ASA/VIRify/MAP V6 pipeline results that were run outside the normal batch-orchestrated production pipeline (e.g. results a person ran manually and now needs registered in MGnify). 

To support this it:

- Generalizes `copy_single_analysis_results() `to work with or without AssemblyAnalysisBatch context.
- Splits `set_post_assembly_analysis_states()` into `update_asa_batch_relations_from_reports()` (batch-related logic) and `update_analyses_from_asa_reports()` (DB objects update logic) to reuse the latter. 
- Extracts `assembly_analysis_results_importer()` and `mark_analyses_with_failed_status()` out of the batch-specific `assembly_analysis_batch_results_importer()` and `process_import_results()` tasks so the same validate→import→mark-failed logic can run over a plain list of Analysis objects.
- Extracts a shared `_run_assembly_summary_generator()` helper and adds `generate_assembly_analysis_pipeline_summary()` (study-level, no batch).
- Relocates shared test fixtures (`assembly_test_scenario`, `test_workspace`, `analyse_study_input_mocker`) into workflows/fixtures/assembly_analysis/conftest.py, and adds unit tests for the new get_study_accession_for_assembly.


ochkalova:
@copilot resolve the merge conflicts in this pull request

workflows/ena_utils/ena_api_requests.py
```diff
@@ -712,3 +712,79 @@ def sync_study_metadata_from_ena(study: ena.models.Study):
...
+            result=ENAPortalResultType.ANALYSIS,
+            fields=[
+                _.ANALYSIS_ACCESSION,
+                _.STUDY_ACCESSION,
+                _.SECONDARY_STUDY_ACCESSION,
+            ],
+            limit=1,
+            query=ENAAnalysisQuery(analysis_accession=assembly_accession),
+            data_portals=[ENAPortalDataPortal.METAGENOME, ENAPortalDataPortal.ENA],
+        ).get()
```
SandyRogers: Is the out-or-prod flow intended to support Private data? This request isn't using the auth at the moment, so it will never succeed if the study is private. This is probably a fine limitation, but we should be aware of it!

requirements-dev.txt
```diff
@@ -8,5 +8,6 @@ pytest-env
 pytest-sugar
 pytest-responses
 pytest-xdist
+pytest-mock
```
SandyRogers, line 11: Any particular reason to introduce pytest-mock's mocker rather than just using unittest.mock directly like elsewhere?

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
```diff
@@ -0,0 +1,600 @@
...
+
+    study_accession = study_accessions.pop()
+
+    # Process the study and its assemblies
+    logger.info(f"Processing study accession: {study_accession}")
+    mgnify_study_id = get_or_create_mgnify_study(study_accession)
+    mgnify_study = analyses.models.Study.objects.get(id=mgnify_study_id)
+    ena_study = ena.models.Study.objects.get_ena_study(study_accession)
+
+    if mgnify_study.is_private:
```
SandyRogers, line 154: I'm not sure we'd get this far if private; does the `get_study_accession_for_assembly` handle private?

workflows/ena_utils/ena_api_requests.py
```diff
@@ -712,3 +712,79 @@ def sync_study_metadata_from_ena(study: ena.models.Study):
...
+            )
+
+        logger.info(f"Assembly {assembly_accession} belongs to study {study_accession}")
+        return study_accession
+
+    except Exception as e:
+        logger.exception(f"Error querying ENA API for assembly {assembly_accession}")
+        raise ValueError(
+            f"Failed to derive study accession for assembly {assembly_accession}: {e}"
+        )
```
mberacochea: Maybe we can just remove this massive try:catch, and let the caller handle this?

Nested code makes it tricky to read IMO (not super important though)

workflows/ena_utils/ena_api_requests.py
```diff
@@ -712,3 +712,79 @@ def sync_study_metadata_from_ena(study: ena.models.Study):
...
+    retries=RETRIES,
+    retry_delay_seconds=RETRY_DELAY,
+    cache_key_fn=task_input_hash,
+    task_run_name="Get study for assembly {assembly_accession} from ENA",
+)
+def get_study_accession_for_assembly(
+    assembly_accession: str,
+) -> str:
+    """
+    Queries the ENA Portal API with the assembly accession (ERZ format) to fetch
```
mberacochea, line 863: @SandyRogers should we care about co-assemblies here?

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
```diff
@@ -0,0 +1,600 @@
...
+    logger.info(
+        "Step 1: Parsing samplesheet structure and extracting assembly accessions..."
+    )
+
+    assembly_accessions = _parse_and_validate_samplesheet(samplesheet_path)
+    logger.info(
+        f"Extracted {len(assembly_accessions)} assembly accessions from samplesheet"
+    )
+    for acc in assembly_accessions:
+        logger.info(f"  - {acc}")
```
mberacochea: (Maybe) This one as an prefect artifact would be useful, for debugging stuff.

If we have the time it may be worth to add it

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
```diff
@@ -0,0 +1,600 @@
...
+
+    if mgnify_study.is_private:
+        logger.info(f"{mgnify_study} is a private study.")
+
+    # Run this function to fetch related runs/samples for this study and save them in the DB
+    get_study_assemblies_from_ena(study_accession, limit=10000)
+
+    # Set results_dir to the provided external path
+    logger.info(f"Set results_dir to {results_path}")
+    mgnify_study.results_dir = str(results_path)
```
mberacochea: So, the results dir of the study should be within the production system. 
In this case, the results dir is a parameter so we should proliferate all sorts of paths in the system.

mberacochea: it maybe that we add a copy step first :brain:

mberacochea (commented):
I haven't finished... but it is looking good :)

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
ochkalova: I removed this code and instead added function argument `results_dir`
[avoid setting study.results_dir to an external path](https://github.com/EBI-Metagenomics/emgapi-v2/pull/250/commits/b5d797b5fc3ec2393d6fa147252eff8349d52c40)

workflows/ena_utils/ena_api_requests.py
ochkalova: [refactor: remove try:exept block from get_study_accession_for_assembly](https://github.com/EBI-Metagenomics/emgapi-v2/pull/250/commits/02a64a033ffa4bac99bec27c226345c14fdc2a5c)

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
ochkalova: Added by Copilot, looks legit to me 😄 
[feat: add create_table_artifact for ingested assembly accessions](https://github.com/EBI-Metagenomics/emgapi-v2/pull/250/commits/17bec63cc9e40a74a087206e087dc99db9881236)

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
```diff
@@ -0,0 +1,604 @@
...
+
+    # Only import if validation passed for at least one analysis
+    successful_validations = [r for r in validation_results if r.success]
+    if successful_validations:
+        logger.info(
+            f"Importing {len(successful_validations)} validated {pipeline_type} analyses..."
+        )
+        import_results = assembly_analysis_results_importer(
+            analyses, schema, base_path, pipeline_type, validation_only=False
+        )
```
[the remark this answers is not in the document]
ochkalova: Already added

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
```diff
@@ -0,0 +1,606 @@
...
+            analyses=analyses,
+            results_workspace=Path(results_dir),
+            destination_root=nfs_results_root,
+            timeout=timeout,
+        )
+    )
+    update_results_dirs_from_copy_results(
+        nfs_copy_results,
+        study_results_dir=nfs_results_root / study_prefix,
+    )
```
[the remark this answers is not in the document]
ochkalova: Resolved in [fix: only copy to production of analysis passed validation](https://github.com/EBI-Metagenomics/emgapi-v2/pull/250/commits/9d1091bf2bcfc88f9283863b97c206747c18f43a)

workflows/ena_utils/ena_api_requests.py
SandyRogers: I don't think they should impact this bit since the assembly is still only in 1 study

workflows/flows/analyse_study_tasks/shared/copy_v6_pipeline_results.py
```diff
@@ -309,11 +324,27 @@ def copy_single_analysis_results(
...
+            return batch.get_pipeline_workspace(pipeline) / assembly_accession
+
+        pipeline_workspace_names = {
+            AssemblyAnalysisPipeline.ASA: "asa",
+            AssemblyAnalysisPipeline.VIRIFY: "virify",
+            AssemblyAnalysisPipeline.MAP: "map",
+        }
+        return (
+            results_workspace / pipeline_workspace_names[pipeline] / assembly_accession
+        )
```
SandyRogers:
```suggestion
        return (
            results_workspace / pipeline.value / assembly_accession
        )
```

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
```diff
@@ -0,0 +1,617 @@
...
+
+    study_prefix = accession_prefix_separated_dir_path(study.first_accession, -3)
+    nfs_results_root = Path(results_dir) / "results"
+
+    external_results_root = (
+        settings.EMG_CONFIG.slurm.private_results_dir
+        if study.is_private
+        else settings.EMG_CONFIG.slurm.ftp_results_dir
+    )
```
SandyRogers:
```suggestion
    external_results_root = (
       EMG_CONFIG.slurm.private_results_dir
        if study.is_private
        else EMG_CONFIG.slurm.ftp_results_dir
    )
```

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
```diff
@@ -0,0 +1,617 @@
+import csv
+from pathlib import Path
+from textwrap import dedent as _
+from typing import List, Optional
+
+from django.conf import settings
```
SandyRogers:
```suggestion
```

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
```diff
@@ -0,0 +1,617 @@
...
+    )
+
+    assembly_accessions = _parse_and_validate_samplesheet(samplesheet_path)
+    logger.info(
+        f"Extracted {len(assembly_accessions)} assembly accessions from samplesheet"
+    )
+
+    create_table_artifact(
+        key="out-of-production-assembly-accessions",
+        table=[{"assembly_accession": acc} for acc in assembly_accessions],
```
SandyRogers, line 111: Be aware that if this is a VERY long list (I guess thousands), this may fail because `create_table_artifact` will sends a huge HTTP request too big for the traffic manager limit.

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
```diff
@@ -0,0 +1,617 @@
...
+    logger.info(f"Ingestion complete for study {mgnify_study.accession}")
+
+
+# ============================================================================
+# HELPER FUNCTIONS
+# ============================================================================
+
+
+@task
+def _parse_and_validate_samplesheet(samplesheet_path: Path) -> list[str]:
```
SandyRogers, line 316: maybe these helper tasks should live in their own files in workflows/flows/analysis/assembly/tasks/ ?

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
```diff
@@ -0,0 +1,617 @@
...
+        )
+
+    return assembly_accessions
+
+
+@task
+def _validate_results_structure(
+    results_path: Path,
+    assemblies_accessions: list[str],
+) -> None:
```
SandyRogers: @mberacochea did you already do a `mgnolia` schema for ASA or did I dream that?

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
```diff
@@ -0,0 +1,617 @@
...
+    analyses_to_update = update_analyses_from_asa_reports(
+        analyses,
+        qc_failed_assemblies,
+        analysed_assemblies,
+        analysed_assemblies_csv,
+        logger=logger,
+    )
+
+    # Bulk update - slightly gentler on the DB
+    Analysis.objects.bulk_update(analyses_to_update, ["status"])
```
SandyRogers: Is there a use case where these may update the status of long-previously created Analyses? Because in this config bulk_update will not change any timestamps on the models (it skips Django .save method), so we might ignore them in timestamp-driven places like the EBI Search dump. This could be fixed by making `update_analyses_from_asa_reports` also set `Analysies.updated_at = now` and then `Analysis.objects.bulk_update(analyses_to_update, ["status", "updated_at])`

workflows/flows/analyse_study_tasks/shared/copy_v6_pipeline_results.py
mberacochea: I think this is not necessary, I would have the 3 directories set in a clean if and be done with this:
```python
asa_source_base = ""
virify_source_base = ""
map_source_base = ""
if has_batch_context:
    asa_source_base = batch.get_pipeline_workspace(AssemblyAnalysisPipeline.ASA) / assembly_accession
    virify_source_base = batch.get_pipeline_workspace(AssemblyAnalysisPipeline.VIRIFY) / assembly_accession
    map_source_base = batch.get_pipeline_workspace(AssemblyAnalysisPipeline.MAP) / assembly_accession
else:
     asa_source_base = results_workspace / AssemblyAnalysisPipeline.ASA.value / assembly_accession
    virify_source_base = results_workspace / AssemblyAnalysisPipeline.VIRIFY.value / assembly_accession
    map_source_base = results_workspace / AssemblyAnalysisPipeline.MAP.value / assembly_accession
```

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
```diff
@@ -0,0 +1,617 @@
...
+    logger.info("Step 2: Validating results structure and consistency...")
+    _validate_results_structure(results_path, assembly_accessions)
+
+    # =========================================================================
+    # STEP 3: Create/fetch Study, Assemblies and related objects, set Biome
+    # =========================================================================
+    logger.info("Step 3: Creating/fetching study and related models...")
+
+    # Find study accession for each assembly accession in the samplesheet
+    study_accessions = set()
```
mberacochea, line 124: This validation should happen before the folder validation... just cheaper and faster.

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
```diff
@@ -0,0 +1,617 @@
+import csv
+from pathlib import Path
+from textwrap import dedent as _
+from typing import List, Optional
+
+from django.conf import settings
+from prefect import flow, get_run_logger, suspend_flow_run, task
```
mberacochea: Import the flows from:
```python
from workflows.prefect_utils.flows_utils import (
    django_db_flow as flow,
    django_db_task as task,
)
```
these have some tweaks to prevent db connections from going stale

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
```diff
@@ -0,0 +1,617 @@
...
+    logger.info(f"Processing study accession: {study_accession}")
+    mgnify_study_id = get_or_create_mgnify_study(study_accession)
+    mgnify_study = analyses.models.Study.objects.get(id=mgnify_study_id)
+    ena_study = ena.models.Study.objects.get_ena_study(study_accession)
+
+    if mgnify_study.is_private:
+        logger.info(f"{mgnify_study} is a private study.")
+
+    # Run this function to fetch related runs/samples for this study and save them in the DB
+    get_study_assemblies_from_ena(study_accession, limit=10000)
```
mberacochea, line 160: We have these hardcoded limits in other places, please add a TODO comment. We should adjust them, or refactor them using a better structure.

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
```diff
@@ -0,0 +1,617 @@
...
+
+        initial_data = {}
+        if not needs_biome:
+            initial_data["biome"] = BiomeChoices[str(mgnify_study.biome.path)]
+
+        ingest_analysis_input: ImportAssemblyAnalysisInput = suspend_flow_run(
+            wait_for_input=ImportAssemblyAnalysisInput.with_initial_data(
+                **initial_data,
+                description=_(f"""\
+                        **External Assembly Analysis Ingestion**
```
mberacochea: Minor, slightly pedantic comment.

We are using 2 names here:
- out-of-production
- external

and in the flow that imports miassembler results I named the flow `import_assemblies_from_filesystem_flow`.

We should follow a naming convention, ideally one that everyone likes :)

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
```diff
@@ -0,0 +1,617 @@
...
+    )
+
+    # =========================================================================
+    # STEP 6: Finalize study and create summaries
+    # =========================================================================
+    logger.info("Step 6: Finalizing study and creating summaries...")
+
+    generate_assembly_analysis_pipeline_summary(
+        study=mgnify_study,
+        asa_workspace=results_path / "asa",
```
mberacochea:
```suggestion
        asa_workspace=results_path / AssemblyAnalysisPipeline.ASA.value,
```
I think

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
```diff
@@ -0,0 +1,617 @@
...
+    logger.info("Added study summaries to downloads")
+
+    copy_v6_study_summaries(mgnify_study.accession)
+    logger.info("Copied v6 study summaries")
+
+    # Update study features
+    mgnify_study.refresh_from_db()
+    mgnify_study.features.has_v6_analyses = mgnify_study.analyses.filter(
+        pipeline_version=analyses.models.Analysis.PipelineVersions.v6, is_ready=True
+    ).exists()
```
mberacochea: I would add a defensive bit here, if `has_v6_analyses` is set to false something went wrong

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
```diff
@@ -0,0 +1,617 @@
...
+            import_results,
+            pipeline_type,
+            validation_only=False,
+        )
+    else:
+        logger.error(f"No {pipeline_type} analyses passed the results validation.")
+
+
+@task
+def set_asa_analysis_states(study_results_dir: Path, analyses: List[Analysis]):
```
mberacochea: Passing "primitives" instead of models is better IMO. Flows and task serialise parameters, and a django model is not the best to serialise if then is going to be used (prefect pickles the data for this).

Also, tasks and flows should use the modified versions. This is to prevent that a long running task or flow renders the connection to the db stale, so we have decided to pass one accessions between tasks/flows and fetch them again before using them.

So here I would ass a `List[int]` and fetch the Analysis before using the models.

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
```diff
@@ -0,0 +1,617 @@
...
+        analysed_assemblies_csv,
+        logger=logger,
+    )
+
+    # Bulk update - slightly gentler on the DB
+    Analysis.objects.bulk_update(analyses_to_update, ["status"])
+
+
+@task
+def copy_out_of_production_assembly_analysis_results(
```
mberacochea: Same here with the Study.

BTW, I understand that some tasks/flows in the codebase don't follow this "convention". We will slowly convert them.

workflows/flows/analysis/assembly/tasks/assembly_analysis_batch_results_importer.py
```diff
@@ -126,27 +125,52 @@ def assembly_analysis_batch_results_importer(
...
 
+    results = assembly_analysis_results_importer(
+        analyses, schema, base_path, pipeline_type, validation_only
+    )
+
+    return results
+
+
+def assembly_analysis_results_importer(
+    analyses: List[Analysis],
```
mberacochea, line 137: This is not a ask so I think that is fine to pass one a list of analysis models.

workflows/flows/analysis/assembly/tasks/assembly_analysis_pipeline_batch_study_summary_generator.py
```diff
@@ -8,13 +8,50 @@
...
 
 
+def _run_assembly_summary_generator(
+    output_dir: Path,
+    asa_workspace: Path,
+    assemblies_csv: Path,
+    output_prefix: str,
+) -> List[Path]:
+
+    logger = get_run_logger()
```
mberacochea, line 26: Does this work? Because this is not a task or flow

workflows/flows/analysis/assembly/tasks/assembly_analysis_pipeline_batch_study_summary_generator.py
```diff
@@ -46,11 +83,45 @@ def generate_assembly_analysis_pipeline_batch_summary(
...
+    return generate_assembly_analysis_pipeline_summary(
+        study=study,
+        asa_workspace=asa_workspace,
+        output_prefix=str(assembly_batch.id),
+    )
+
+
+@task
+def generate_assembly_analysis_pipeline_summary(
+    study: Study,
```
mberacochea: I would send the study accession here

workflows/flows/analysis/assembly/tasks/process_import_results.py
```diff
@@ -85,19 +129,7 @@ def process_import_results(
...
-                save=False,  # Batch will be saved once after all errors logged
-            )
-
+    # Bulk update all analyses at once
     if analyses_to_update:
         Analysis.objects.bulk_update(analyses_to_update, ["status"])
         logger.info(f"Marked {len(analyses_to_update)} analyses as ANALYSIS_QC_FAILED")
-
-    # Persist the errors in the batch
-    if failed_analysis_ids:
```
mberacochea, line 102: Why was this removed?

workflows/ena_utils/ena_api_requests.py
```diff
@@ -710,3 +710,72 @@ def sync_study_metadata_from_ena(study: ena.models.Study):
...
         study.save()
+
+
+@task(
+    retries=RETRIES,
+    retry_delay_seconds=RETRY_DELAY,
+    cache_key_fn=task_input_hash,
+    task_run_name="Get study for assembly {assembly_accession} from ENA",
+)
+def get_study_accession_for_assembly(
```
mberacochea, line 859: This method should handle private studies too. Checkout `get_study_assemblies_from_ena`, it has a `study.is_private` bit to then provide the ena_auth credentials

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
```diff
@@ -0,0 +1,617 @@
...
+
+    generate_assembly_analysis_pipeline_summary(
+        study=mgnify_study,
+        asa_workspace=results_path / "asa",
+    )
+
+    add_assembly_study_summaries_to_downloads(mgnify_study.accession)
+    logger.info("Added study summaries to downloads")
+
+    copy_v6_study_summaries(mgnify_study.accession)
```
mberacochea: :robot: This one should only copy ASA summaries

```suggestion
    copy_v6_study_summaries(mgnify_study.accession, analysis_type=AnalysisType.ASSEMBLY)
```

workflows/flows/analyse_study_tasks/shared/copy_v6_pipeline_results.py
```diff
@@ -271,25 +271,40 @@ def copy_assembly_batch_results_to_destination_folder(
 @task
 def copy_single_analysis_results(
```
mberacochea, line 272: A suggestion the robots made, which I _kind_ of agree, is to have another copy task for the non-batch ones. A bit of duplicated code but I find it easier to read:
```python
@task
def copy_out_of_production_analysis_results(
    analysis: Analysis,
    destination_root: Path,
    results_workspace: Path,
    timeout: int = 14400,
) -> BatchCopyResult:
    logger = get_run_logger()

    assembly_accession = analysis.assembly_or_run.first_accession
    study_accession = analysis.study.first_accession

    destination_base = (
        Path(destination_root)
        / accession_prefix_separated_dir_path(study_accession, -3)
        / accession_prefix_separated_dir_path(assembly_accession, -3)
        / analysis.pipeline_version
        / Analysis.ExperimentTypes.ASSEMBLY.label.lower()
    )

    copy_errors: list[CopyError] = []

    # ASA is required.
    asa_source = results_workspace / "asa" / assembly_accession
    asa_success, asa_errors = copy_schema_directories(
        schema=AssemblyResultSchema(),
        source_base=asa_source,
        destination_base=destination_base,
        timeout=timeout,
    )
    copy_errors.extend(asa_errors)

    if not asa_success:
        logger.error(
            f"ASA copy for {analysis.accession} failed"
        )

    # VIRify and MAP are optional.
    optional_pipelines = (
        ("virify", "VIRify", VirifyResultSchema),
        ("map", "MAP", MapResultSchema),
    )

    for directory_name, pipeline_name, schema_class in optional_pipelines:
        source_base = (
            results_workspace
            / directory_name
            / assembly_accession
        )

        if not source_base.exists():
            logger.info(
                f"No {pipeline_name} output found for {analysis.accession} "
                f"at {source_base}; skipping optional copy"
            )
            continue

        success, errors = copy_schema_directories(
            schema=schema_class(),
            source_base=source_base,
            destination_base=destination_base,
            timeout=timeout,
        )
        copy_errors.extend(errors)

        if not success:
            logger.error(
                f"{pipeline_name} copy for {analysis.accession} failed"
            )

    if copy_errors:
        return BatchCopyResult(
            analysis_id=analysis.id,
            destination_folder=destination_base,
            success=False,
            errors=copy_errors,
        )

    logger.info(
        f"Analysis {analysis.accession} results copied to {destination_base}"
    )

    return BatchCopyResult(
        analysis_id=analysis.id,
        destination_folder=destination_base,
        success=True,
    )
```

workflows/flows/analysis/assembly/tasks/assembly_analysis_pipeline_batch_study_summary_generator.py
ochkalova: Resolved in [pass primitives (e.g. List[int] of ids), not Django model instances, …](https://github.com/EBI-Metagenomics/emgapi-v2/pull/250/commits/08009cf93ffe1c297a97dc8c4e6336fce170a32a)

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
ochkalova: Resolved in [pass primitives (e.g. List[int] of ids), not Django model instances, …](https://github.com/EBI-Metagenomics/emgapi-v2/pull/250/commits/08009cf93ffe1c297a97dc8c4e6336fce170a32a)

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
ochkalova: Resolved in [pass primitives (e.g. List[int] of ids), not Django model instances, …](https://github.com/EBI-Metagenomics/emgapi-v2/pull/250/commits/08009cf93ffe1c297a97dc8c4e6336fce170a32a)

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
ochkalova: I agree, I renamed this flow to "out-of-production" at some point (Sandy's idea). I think it's the most straightforward and clear naming
The line you pointed at is an artifact from initial naming and has been already fixed in [clarify pop-up description message](https://github.com/EBI-Metagenomics/emgapi-v2/pull/250/commits/7ec0dd553e1c9b8cca9199894a8a4125c5a007b7)

If you want sonething unified, I vote for "out-of-production". "external" is used in the code to refer to filesystem where we store results and "from_filesystem" is too vague IMHO

workflows/flows/analyse_study_tasks/shared/copy_v6_pipeline_results.py
ochkalova: Makes sense. I changed in [separate copying of assembly analysis results with/without batch cont…](https://github.com/EBI-Metagenomics/emgapi-v2/pull/250/commits/0ca33b72952a081bbd444ddf279a8a450fd467fe)

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
ochkalova: Fixed in [import flow and task from workflows.prefect_utils.flows_utils](https://github.com/EBI-Metagenomics/emgapi-v2/pull/250/commits/e450f71a3ee7ba14f0a015877d4a049d5ed278f7)

workflows/flows/analyse_study_tasks/shared/copy_v6_pipeline_results.py
ochkalova: Irrelevant after https://github.com/EBI-Metagenomics/emgapi-v2/pull/250#discussion_r3756328945

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
ochkalova: Done!

workflows/flows/analysis/assembly/tasks/process_import_results.py
ochkalova: It wasn't. The diff just looks misleading. This code was just moved back up into process_import_results itself (lines 64–76 in the current file)

workflows/flows/analysis/assembly/tasks/assembly_analysis_pipeline_batch_study_summary_generator.py
ochkalova: AFAIU it should work fine. It will find Prefect's run context in the parent function

workflows/ena_utils/ena_api_requests.py
ochkalova: Added in [Add private study support to get_study_accession_for_assembly](https://github.com/EBI-Metagenomics/emgapi-v2/pull/250/commits/ff9a20c276faee7a53704536156c056c218c1f9e)

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
ochkalova: Moved up in [move study-accession/single-study validation before filesystem valida…](https://github.com/EBI-Metagenomics/emgapi-v2/pull/250/commits/b052286c0be6bf9a8a6f13f405b03f13feab4d0e)

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
ochkalova: Looks reasonable, I have removed all literals "asa" "virify" and "map" from the flow's code 
[replace literal pipeline directory strings with AssemblyAnalysisPipel…](https://github.com/EBI-Metagenomics/emgapi-v2/pull/250/commits/2345ff992b8656db6babd5b8eb44ee1d6c550f88)

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
ochkalova: Good catch, Martin! You helped to identify another bug 🫠
This whole check should happen after copying of results, not before...
Fixed in [fail loudly if study ends up with no ready v6 analyses after import](https://github.com/EBI-Metagenomics/emgapi-v2/pull/250/commits/91dc916802cc1b5028e37356c3acdfa89330bba7)

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
mberacochea: Alright, I think out-of-production is a good one too. I'll rename the assemblies importer one :)

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
ochkalova: And after I found more functions that take Django objects and updated them as well: [again: pass primitives (e.g. List[int] of ids), not Django model inst…](https://github.com/EBI-Metagenomics/emgapi-v2/pull/250/commits/9d5eddbd05cc68adb6cca40e7d135197e67d4ae7)

workflows/ena_utils/ena_api_requests.py
```diff
@@ -710,3 +710,84 @@ def sync_study_metadata_from_ena(study: ena.models.Study):
...
+        query=ENAAnalysisQuery(analysis_accession=assembly_accession),
+        data_portals=[ENAPortalDataPortal.METAGENOME, ENAPortalDataPortal.ENA],
+    )
+
+    # We don't yet know which study this assembly belongs to, so we can't look up
+    # is_private from our own DB.
+    # Try publicly first, and only fall back to authenticated access if that finds
+    # nothing - the assembly may belong to a private study.
+    portal_analyses = analysis_request.get(raise_on_empty=False)
+    if not portal_analyses:
```
mberacochea, line 905: One tiny bit, add the public-first-private-fallback to the docstring

workflows/flows/analyse_study_tasks/shared/copy_v6_pipeline_results.py
```diff
@@ -411,6 +411,108 @@ def copy_single_analysis_results(
...
+        results_workspace / AssemblyAnalysisPipeline.ASA.value / assembly_accession
+    )
+    asa_copy_success, asa_copy_errors = copy_schema_directories(
+        schema=AssemblyResultSchema(),
+        source_base=asa_source_base,
+        destination_base=destination_base,
+        timeout=timeout,
+    )
+    if not asa_copy_success:
+        logger.error(f"ASA {analysis.accession} copy for {assembly_accession} failed")
```
mberacochea: If there are errors, should we stop?

workflows/flows/analysis/assembly/tasks/assembly_analysis_pipeline_batch_study_summary_generator.py
mberacochea: Cool cool

workflows/flows/analysis/assembly/tasks/set_post_assembly_analysis_states.py
mberacochea, line 1: Please ask the robots to add docstrings to these methods.

workflows/tests/test_import_out_of_production_assembly_analysis_results.py
```diff
@@ -0,0 +1,593 @@
...
+            study=raw_reads_mgnify_study,
+            sample=raw_reads_mgnify_sample[0],
+            ena_study=raw_reads_mgnify_study.ena_study,
+            assembly=mgnify_assemblies[0],
+            pipeline_version="6.0",
+        )
+        return analysis
+
+    @patch(
+        "workflows.flows.analysis.assembly.flows.import_out_of_production_assembly_analysis_results.copy_single_out_of_production_analysis_results"
```
mberacochea: Looking at the package name:
```
workflows.flows.analysis.assembly.flows.import_out_of_production_assembly_analysis_results.
```

it occurred to me that instead of long names: `import_out_of_production_assembly_analys` we should create a python module for `out_of_production` instead.

Like this:
```
workflows.flows.analysis.assembly.flows.out_of_production.import_assembly_analysis_results

out_of_production\
  import_assembly_analysis.py
  import_miassembler.py
```

Anyway, not for this PR.

workflows/tests/test_import_out_of_production_assembly_analysis_results.py
```diff
@@ -0,0 +1,593 @@
...
+
+
+def create_input_fasta_files(fasta_dir: Path, assembly_accessions: List[str]):
+    """Create dummy input fasta files for given assembly accessions."""
+    for acc in assembly_accessions:
+        fasta_path = fasta_dir / f"{acc}.fa.gz"
+        fasta_path.parent.mkdir(parents=True, exist_ok=True)
+        with fasta_path.open("w") as f:
+            f.write(f">Dummy contig for {acc}\n")
+            f.write("ATGC" * 10 + "\n")
```
mberacochea, line 186: Image

workflows/tests/test_import_out_of_production_assembly_analysis_results.py
ochkalova: Nice. It will be tidy and organised. Just like we all love 😍

workflows/tests/test_import_out_of_production_assembly_analysis_results.py
ochkalova: 😆

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
mberacochea: Yes, I forgot to comment this. I've bumped into this often. I think for this one... it is probably OK

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
mberacochea: Yes, and it does it this further down. This is a quick validation on the base folders to check that it has the asa, map and virify folders in it. Then schemas are used further down

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
mberacochea: This is a good point. Also, the code should act defensibly if an analysis already exists for a combination of pipeline version + assembly accession... otherwise we could insert duplicates.

requirements-dev.txt
ochkalova: TBH I don't remember... I think at some point it just seemed more convenient 🤔

workflows/ena_utils/ena_api_requests.py
ochkalova: Nice catch, Martin also spotted this and it was resolved in [Add private study support to get_study_accession_for_assembly](https://github.com/EBI-Metagenomics/emgapi-v2/pull/250/commits/ff9a20c276faee7a53704536156c056c218c1f9e)

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
ochkalova: Yes, it was added in [Add private study support to get_study_accession_for_assembly](https://github.com/EBI-Metagenomics/emgapi-v2/pull/250/commits/ff9a20c276faee7a53704536156c056c218c1f9e)

workflows/flows/analyse_study_tasks/shared/copy_v6_pipeline_results.py
ochkalova: Great refactoring, but the code was updated already 😞

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
ochkalova: I will leave it for now, because after testing some changes will probably have to be made anyway.
If this artifact won't be helpful, I will remove it

workflows/ena_utils/ena_api_requests.py
ochkalova: Added the public-first/private-fallback note to the docstring.

workflows/flows/analysis/assembly/tasks/set_post_assembly_analysis_states.py
ochkalova: Added docstrings to `_read_report_csv`, `parse_asa_end_of_run_reports`, `update_asa_batch_relations_from_reports`, and `update_analyses_from_asa_reports`.

requirements-dev.txt
ochkalova: I found why I started to use it:
The test `test_import_out_of_production_assembly_analysis_results_real_data` is very big and it has 9 pytest fixtures. Because `@patch` decorators inject mocks positionally, stacked bottom-up, ahead of the named fixture args, mixing multiple stacked `@patch` decorators with that many named pytest fixtures is a pain. At some point I got stuck in this ordering/readability trap and decided that I need to get rid of it...

Also it's more convenient because mocker.patch(...) allows to use callables that are defined mid-test. Overall, it seems like newer and better solution. If you don't insist, I would prefer to keep it.

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
ochkalova: After some thinking I decided that this one in particular is overdecorated and @task can be removed ([from _validate_results_structure and _parse_and_validate…](https://github.com/EBI-Metagenomics/emgapi-v2/pull/250/commits/73593dc243f10c9c7aaffa3cee87b28fb3f46411)). As for others, I moved them to workflows/flows/analysis/assembly/tasks/ in
[move tasks from import_out_of_production_assembly_analysis_results.py…](https://github.com/EBI-Metagenomics/emgapi-v2/pull/250/commits/edc038d2b6cc3289dde1f8042dff62e42d91d338)

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
ochkalova: I've added fix that Sandy suggested for bulk_update() to this function and to `set_post_assembly_analysis_states()` that is used in ASA automation
[manually set "updated_at" before bulk_update()](https://github.com/EBI-Metagenomics/emgapi-v2/pull/250/commits/51b6faf641f1e30749d1609c8a249ae77559519c)

UPD: found one more case of this bug in another function, fixed in [manually set "updated_at" before bulk_update() in mark_analyses_with_failed_status](https://github.com/EBI-Metagenomics/emgapi-v2/pull/250/commits/9a44a2cea4d58c70d3e989d709efbfd45a582739)

ochkalova: @mberacochea regarding your comment, it seems to me that 
```
analysis, created = analyses.models.Analysis.objects.get_or_create(
    study=study,
    sample=assembly.sample,
    assembly=assembly,
    ena_study=study.ena_study,
    pipeline_version=pipeline,
)
```
is already defensive, it shouldn't produce duplicates for the same  pipeline version + assembly accession

workflows/flows/analyse_study_tasks/shared/copy_v6_pipeline_results.py
ochkalova: I'm not sure, Martin. I used `copy_single_analysis_results` as an example when created this. `copy_single_analysis_results` doesn't stop. From the other side it has a check before copying starts: `batch_analysis_job.asa_status == COMPLETED`, so maybe it's justified... Do you think we need to revise the behaviour in both places?

ochkalova: I've added early exit in [exit early if ASA copying failed](https://github.com/EBI-Metagenomics/emgapi-v2/pull/250/commits/11db5201ecc8ac6c88fbb7e56e2343c889aaea70)

requirements-dev.txt
SandyRogers: Ah nice, good to know!

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
SandyRogers: okay!

workflows/flows/analysis/assembly/flows/import_out_of_production_assembly_analysis_results.py
SandyRogers: Thank you @ochkalova

PR #302, opened 2026-04-01
Feature/sentry

MGS-sails:
This PR:
* Adds Sentry to aid in debugging

---

#### Checklist


emgapiv2/settings.py
```diff
@@ -43,6 +44,11 @@
 
 DEBUG = os.getenv("DJANGO_DEBUG", False)
 
+sentry_sdk.init(
+    dsn=os.getenv("SENTRY_DSN"),
```
SandyRogers: can we use the config.py file for this, rather than an env var? just because it gives more options about how to set it (everything supported by Pydantic Settings)

SandyRogers (changes requested):
Thanks @MGS-sails this will be v helpful!
Only very minor changes requested below.

emgapiv2/settings.py
```diff
@@ -65,6 +66,12 @@ def show_toolbar(request):
 if BASE_URL:
     BASE_URL += "/"
 
+sentry_sdk.init(
```
mberacochea, line 69: I think having an if guard here would be good, to avoid crashes when running this locally for example
```python
if EMG_CONFIG.sentry_dsn:
   sentry_sdk.init...
```

mberacochea (approved):
One tiny comment

SandyRogers (approved):

PR #313, opened 2026-04-10
Wrappers for flow and task - that close stale connections before running

mberacochea:
Wrappers for flow and task - that close stale connections before running
After much tinkering I came up with this approach. There are two helpers now
django_flow and django_tasks, these run a clean stale connections before running
this should help with long running flows (that have many tasks) as before running
tasks or sub-flows connections should be cleared.

I'm open to suggestions :).

---

#### Checklist


workflows/prefect_utils/db_utils.py
mberacochea, line 1: Arg.. I forgot to remove this one

workflows/prefect_utils/flows_utils.py
```diff
@@ -0,0 +1,81 @@
...
+    corrupt the transaction, and because Django's AUTOCOMMIT setting (True by
+    default) mismatches the live autocommit state (False inside atomic()),
+    causing close_if_unusable_or_obsolete() to treat it as a state change and
+    close the connection. Transactions should never span multiple tasks or
+    flows, so in normal operation this guard only triggers in pytest-django,
+    which wraps each test in transaction.atomic() to allow rollback.
+    """
+    for conn in connections.all():
+        if not conn.in_atomic_block:
+            conn.close_if_unusable_or_obsolete()
```
SandyRogers, line 23: Following the comment above, should we have an else here then, which asserts that we are in a pytest environment and if not, raises a logging warning that we are leaving a conn uncleaned in an unexpected (prod) case?

workflows/prefect_utils/flows_utils.py
```diff
@@ -0,0 +1,81 @@
...
+        def wrapper(*args, **kwargs):
+            _close_stale_connections()
+            return fn(*args, **kwargs)
+
+        return prefect_decorator(wrapper)
+
+    return decorator
+
+
+def django_task(**task_kwargs) -> Callable:
```
SandyRogers: does this still allow the handy prefect testing approach, of doing `my_task.fn()` to avoid using the prefect test harness? or does the additional wrapping maybe that impossible? I think it is probably okay just looking at it, but just check... given the changes to tests below I feel maybe it does break something.

SandyRogers (approved):
Thanks @mberacochea . Left a couple of small queries in line... 
Bigger things, maybe just change the imports to use `import django_flow as flow` at the top of each file instead of changing every single decoration. The main reason is just so that we could make a little pre-commit checker to test that every @task/@flow import actually uses the utils version instead of the prefect one.

workflows/prefect_utils/flows_utils.py
mberacochea: I'm not sure, I'll check

workflows/prefect_utils/flows_utils.py
mberacochea: Good point, I'll add this. I mean, having a transaction that spans several tasks (or flows)... is a bad idea.

PR #320, opened 2026-04-13
Add a flow to sync the metadata for the studies

mberacochea:
This PR:
* Using the command sync_studies_with_ena - adds a flow to run this instead of a command

---

#### Checklist

### DIFF
--- a/workflows/flows/housekeeping/sync_studies_with_ena.py
+++ b/workflows/flows/housekeeping/sync_studies_with_ena.py
@@ -1,0 +1,81 @@
+from prefect import get_run_logger
+
+from activate_django_first import EMG_CONFIG  # noqa: F401
+
+import ena.models
+from workflows.ena_utils.ena_api_requests import sync_study_metadata_from_ena
+from workflows.prefect_utils.flows_utils import (
+    django_db_flow as flow,
+    django_db_task as task,
+)
+
+
+@task(name="Sync batch of studies metadata from ENA")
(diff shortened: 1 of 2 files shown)


workflows/flows/housekeeping/sync_studies_with_ena.py
```diff
@@ -0,0 +1,81 @@
...
+
+    Studies are processed in batches to avoid long-running DB connections.
+
+    :param accessions: List of study accessions to sync.
+    :param all_studies: If True, sync all studies.
+    :param batch_size: Number of studies to process per batch (default 50).
+    """
+    logger = get_run_logger()
+
+    if accessions and all_studies:
```
MGS-sails, line 53: Could make this slightly flatter: 

`if accessions and all_studies:
    raise ValueError("Cannot provide both accessions and all_studies")
if not accessions and not all_studies:
    raise ValueError("Must provide either accessions or all_studies=True")

study_accessions = accessions or list(
    ena.models.Study.objects.values_list("accession", flat=True)
)`

MGS-sails (approved):
Thanks @mberacochea . 
Left one comment

workflows/flows/housekeeping/sync_studies_with_ena.py
mberacochea: yes, sorry I didn't see the comment before merging (which only appeared after reloading the side...)

PR #322, opened 2026-04-14
EMG-8989 Adjust "pipeline version undefined" in study summaries

SandyRogers:
This PR:
* adds the pipeline version to the `download_group` of V6 study summaries. It was there in fixtures, but not set when the studies were added.
* this is done in two ways because ASA does them differently to AMP and RAWR
* adds a management command to migrate existing downloads to the new format

This fixes EMG-8989:
* the `.v6.` part of the `download_group` is used by the website to group a study's summaries into their pipeline versions

---

#### Checklist


mberacochea (approved):
Thanks @SandyRogers. It looks sensible to me

PR #327, opened 2026-04-22
Don't persist {asa,virify,map}_results - we don't need them

mberacochea:
The run_batch flows handle re-executions, so we don't need prefect to handle that. I bumped in to the case where I needed to re-run the asa flow to get the results ingested (because the flow died after nextflow ran but it could'nt save the orchestrated job in the db - timeout). So the flow was marked as done but the results never imported. I re-run the run_assembly_batch thingy but prefect refused to do so, as the previuos flow as marked as done. This shouldn't really happen, as we shouldn't get more timeouts now but it could happen again the future (db connections and stuff).

---

#### Checklist


mberacochea:
This just adds code smells to the codebase.

PR #328, opened 2026-04-22
EMG-9743 Support mixed version pipelines, e.g. v6.1 for Amplicon

SandyRogers:
This PR:
* adds supported analysis pipeline versions for all legacy released version (v1.0 - v5.0)
* adds support for a v6 version family, removing some assumptions that all v6 pipelines would be in-sync.
  * this is because amplicon will shortly be v6.1 whilst raw-reads and assembly-analysis stay on v6.0

Codex-assisted.

---

#### Checklist

### DIFF
--- a/analyses/models.py
+++ b/analyses/models.py
@@ -930,16 +930,10 @@ class Analysis(
     def pipeline_version_from_config(
         cls, pipeline_version: str
     ) -> "Analysis.PipelineVersions":
-        normalized = pipeline_version.strip().lower()
-        version_map = {
-            cls.PipelineVersions.v6.label.lower(): cls.PipelineVersions.v6.value,
-            "v6": cls.PipelineVersions.v6.value,
-            cls.PipelineVersions.v6_1.label.lower(): cls.PipelineVersions.v6_1.value,
-            "v6.1": cls.PipelineVersions.v6_1.value,
-        }
+        normalized = pipeline_version.strip().lower().rstrip(".0")
         try:
-            return cls.PipelineVersions(version_map[normalized])
-        except KeyError as exc:
+            return getattr(cls.PipelineVersions, normalized)
+        except AttributeError as exc:
             raise ValueError(
                 f"Unsupported pipeline version {pipeline_version!r}"
             ) from exc
--- a/analyses/models.py
+++ b/analyses/models.py
@@ -913,12 +913,36 @@ class Analysis(
     metadata = models.JSONField(default=dict, blank=True)
 
     class PipelineVersions(models.TextChoices):
+        v1 = "V1", "v1.0"
+        v2 = "V2", "v2.0"
+        v3 = "V3", "v3.0"
+        v4 = "V4", "v4.0"
+        v4_1 = "V4.1", "v4.1"
         v5 = "V5", "v5.0"
         v6 = "V6", "v6.0"
+        v6_1 = "V6.1", "v6.1"
 
     pipeline_version = models.CharField(
         choices=PipelineVersions, max_length=5, default=PipelineVersions.v6
     )
+
+    @classmethod
+    def pipeline_version_from_config(
+        cls, pipeline_version: str
+    ) -> "Analysis.PipelineVersions":
+        normalized = pipeline_version.strip().lower()
+        version_map = {
+            cls.PipelineVersions.v6.label.lower(): cls.PipelineVersions.v6.value,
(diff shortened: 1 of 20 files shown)


mberacochea (approved):
It looks sensible. We definitly need to bring all the pipelines to the same abstractions.. it is getting hard to maintain.

workflows/flows/analysis/pipeline_versions.py
```diff
@@ -0,0 +1,44 @@
...
+    Analysis.ExperimentTypes.METATRANSCRIPTOMIC: EMG_CONFIG.rawreads_pipeline,
+    Analysis.ExperimentTypes.ASSEMBLY: EMG_CONFIG.assembly_analysis_pipeline,
+    Analysis.ExperimentTypes.HYBRID_ASSEMBLY: EMG_CONFIG.assembly_analysis_pipeline,
+    Analysis.ExperimentTypes.LONG_READ_ASSEMBLY: EMG_CONFIG.assembly_analysis_pipeline,
+}
+
+
+def get_pipeline_config_for_experiment_type(
+    experiment_type: Analysis.ExperimentTypes | str,
+):
```
timrozday-mgnify: Return type missing
```suggestion
) -> MGnifyPipelineConfig:
```

workflows/flows/analysis/pipeline_versions.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def get_current_pipeline_version_for_experiment_type(
+    experiment_type: Analysis.ExperimentTypes | str,
+) -> Analysis.PipelineVersions:
+    pipeline_config = get_pipeline_config_for_experiment_type(experiment_type)
+    return Analysis.pipeline_version_from_config(pipeline_config.pipeline_version)
+
+
+def get_v6_family_pipeline_versions_for_experiment_type(
```
timrozday-mgnify, line 34: Could we make this function a little more general? With something like:
```python
def get_configured_family_pipeline_versions_for_experiment_type(
    experiment_type: Analysis.ExperimentTypes | str,
) -> tuple[Analysis.PipelineVersions, ...]:
    
    def _major_version(version_str: str) -> str:
        return version_str.lstrip("vV").split(".", 1)[0]
      
    pipeline_config = get_pipeline_config_for_experiment_type(experiment_type)
    configured_major = _major_version(pipeline_config.pipeline_version)
    return tuple(
        version
        for version in Analysis.PipelineVersions
        if _major_version(version.label) == configured_major
    )
```

analyses/models.py
```diff
@@ -913,13 +913,37 @@ class KnownMetadataKeys:
...
         v5 = "V5", "v5.0"
         v6 = "V6", "v6.0"
+        v6_1 = "V6.1", "v6.1"
 
     pipeline_version = models.CharField(
         choices=PipelineVersions, max_length=5, default=PipelineVersions.v6
     )
 
+    @classmethod
+    def pipeline_version_from_config(
```
timrozday-mgnify, line 930: Can this be more general to support future pipeline versions? Could generate a `dict` from the version `enums` like:
```python
def pipeline_version_from_config(version_str: str) -> Analysis.PipelineVersions:
    normalized = version_str.strip().lower()
    aliases: dict[str, Analysis.PipelineVersions] = {}
    for version in Analysis.PipelineVersions:
        label = version.label.lower()
        aliases[label] = version
        if label.endswith(".0"):
            aliases.setdefault(label[:-2], version)  # "v6.0" -> also "v6"
    try:
        return aliases[normalized]
    except KeyError as exc:
        raise ValueError(f"Unsupported pipeline version {version_str!r}") from exc
```

timrozday-mgnify (changes requested):
Thanks @SandyRogers , didn't realise a version bump would be such an issue!

I've tried to suggest ways to make these changes more general for future version changes (well, I asked the robots).

It would be nice to role out similar such changes to the other analysis pipelines in a future PR (I can help with that?)

workflows/flows/analysis/pipeline_versions.py
SandyRogers: Thanks Tim – good suggestion. I have decided to leave as is, as the v6-specific stuff is really a temporary "hack" to support the v6-early-data-release website, until the legacy data are fully imported. Have added a TODO to make sure we clean this up, instead of generalising the logic which won't be needed afterwards.

PR #330, opened 2026-04-23
Sync samples with ena flow - WIP

mberacochea:
This PR: add a flow to sync the samples metadata from ENA

---

#### Checklist

### DIFF
--- a/workflows/flows/housekeeping/sync_samples_with_ena.py
+++ b/workflows/flows/housekeeping/sync_samples_with_ena.py
@@ -1,0 +1,84 @@
+from prefect import get_run_logger
+
+from activate_django_first import EMG_CONFIG  # noqa: F401
+
+import ena.models
+from workflows.ena_utils.ena_api_requests import sync_sample_metadata_from_ena
+from workflows.prefect_utils.flows_utils import (
+    django_db_flow as flow,
+    django_db_task as task,
+)
+
+
+@task(name="Sync batch of samples metadata from ENA")
+def sync_samples(sample_accessions: list[str]) -> list[str]:
+    """Sync metadata for a batch of samples from ENA.
+
+    Each sample is synced individually with try/except so that a failure
+    for one sample does not block the rest of the batch.
+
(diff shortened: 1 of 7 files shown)


workflows/flows/housekeeping/sync_samples_with_ena.py
```diff
@@ -0,0 +1,84 @@
...
+        sample_accessions = list(
+            ena.models.Sample.objects.values_list("accession", flat=True)
+        )
+
+    total = len(sample_accessions)
+    logger.info(f"Syncing metadata for {total} samples in batches of {batch_size}")
+
+    all_failed = []
+    for i in range(0, total, batch_size):
+        batch = sample_accessions[i : i + batch_size]
```
SandyRogers, line 77: A more efficient way to batch would be to iterate through studies, and modify/duplicate `sync_sample_metadata_from_ena` to accept a study idea and query the ENA endpoint for samples but queried by study id, to minimise the total number of requests when fetching all.

workflows/flows/housekeeping/sync_samples_with_ena.py
```diff
@@ -0,0 +1,84 @@
...
+        logger.info(f"Processing batch {i // batch_size + 1} ({len(batch)} samples)")
+        failed = sync_samples(batch)
+        all_failed.extend(failed)
+
+    if all_failed:
+        logger.warning(
+            f"Failed to sync {len(all_failed)} samples: {', '.join(all_failed)}"
+        )
```
SandyRogers, line 92: I think it would be better to make a table Artefact of the failed ones rather than print a potentially very long log line (e.g. if ENA Portal is down, this list could be 100,000 accessions long)

SandyRogers (commented):
Just 2 suggestions, neither critical.

workflows/flows/housekeeping/sync_samples_with_ena.py
mberacochea: yes, good point. I shall ask the robots to do this

PR #332, opened 2026-04-24
Feature/cobs search

MGS-sails:
This PR: incorporates the genome-search endpoint. 
This endpoint enables the gene search (COBS) feature from the frontend to work. 
The endpoint receives the gene search request, both in JSON and Form content-type, parses them, and sends it off to the external Cobs search service.

*

---

#### Checklist


emgapiv2/settings.py
```diff
@@ -407,3 +407,10 @@ def get_environment(request):
 # Django Nginx Secure Links settings - pre-signed URLs for private data
 SECURE_LINK_SECRET_KEY = os.getenv("PRIVATE_DATA_SECURE_LINK_SECRET_KEY", SECRET_KEY)
 SECURE_LINK_EXPIRATION_SECONDS = 86400
+
+
+# Genome search backend proxy
+GENOME_SEARCH_PROXY = os.getenv(
+    "GENOME_SEARCH_PROXY",
+    "https://cobs-genome-search-01.mgnify.org/search",
+)
```
SandyRogers: This would be better in `emgapiv2/config.py:ServiceURLsConfig` to follow the code style guide

emgapiv2/urls.py
```diff
@@ -48,6 +48,9 @@
     path(f"{BASE_URL}__debug__/", include("debug_toolbar.urls")),
     path(f"{BASE_URL}fieldfiles/", include("db_file_storage.urls")),
     path(BASE_URL, api.urls),
+    # Provide a stable '/api/' alias for local and default setups where BASE_URL is empty
+    # Avoid double-mounting if BASE_URL is already 'api/'
+    # path("api/", api.urls),
```
SandyRogers: Hmm, unsure what this is getting at?

emgapiv2/api/genome_search.py
```diff
@@ -0,0 +1,213 @@
...
+class GenomeFragmentSearchOut(Schema):
+    data: GenomeSearchData
+
+
+def _backend_url() -> str:
+    return getattr(
+        settings,
+        "GENOME_SEARCH_PROXY",
+        "https://cobs-genome-search-01.mgnify.org/search",
+    )
```
SandyRogers: Use EMG_CONFIG here

emgapiv2/api/genome_search.py
```diff
@@ -0,0 +1,213 @@
+import logging
+from typing import List, Optional, Dict, Any, Tuple
+import json
+from json import JSONDecodeError
+
+import requests
```
SandyRogers: If possible use `httpx` – most of the codebase is on that instead of `requests`. Doesn't make much difference to this implementation as it stands, but for future changes like async, global outbound retry logic etc it is handy to have everything using the same lib

emgapiv2/api/genome_search.py
```diff
@@ -0,0 +1,213 @@
...
+    """
+    content_type = request.headers.get("content-type", "").lower()
+    is_form = (
+        content_type.startswith("multipart/")
+        or content_type.startswith("application/x-www-form-urlencoded")
+        or (
+            hasattr(request, "POST")
+            and (bool(request.POST) or bool(getattr(request, "FILES", None)))
+        )
+    )
```
SandyRogers, line 69: In practice is this needed above a simpler

```suggestion
    is_form = bool(request.POST) or bool(request.FILES)
```

?

SandyRogers (changes requested):
Nice, thanks @MGS-sails ! Just a couple of small changes requested

emgapiv2/api/genome_search.py
MGS-sails: I think the content_type check is still necessary since requests with JSON body would also still be a post.

emgapiv2/api/genome_search.py
MGS-sails: changed to httpx

emgapiv2/api/genome_search.py
SandyRogers: Ah yes, I'm not sure what I was thinking.

emgapiv2/settings.py
```diff
@@ -407,3 +407,7 @@ def get_environment(request):
 # Django Nginx Secure Links settings - pre-signed URLs for private data
 SECURE_LINK_SECRET_KEY = os.getenv("PRIVATE_DATA_SECURE_LINK_SECRET_KEY", SECRET_KEY)
 SECURE_LINK_EXPIRATION_SECONDS = 86400
+
+
+# Genome search backend proxy
+GENOME_SEARCH_PROXY = EMG_CONFIG.service_urls.genome_search_proxy
```
SandyRogers: I think this is now unused? Since in `genome_search.py` you get it (correctly) via `EMG_CONFIG`?

SandyRogers (approved):
Just one unused (I think) variable otherwise :shipit:

PR #334, opened 2026-04-27
use canonical (pk) for admin pages that can also be accessed by accession

SandyRogers:
This PR:
* makes Study admin page redirect to URL with pk in (instead of accession) when accessed by the latter
* this fixes a bug where navigating to Study change page from a related item that references the Study *by accession* in its fk, led to broken Study Reports sice unfold assumed the object_id will be the pk

---

#### Checklist


mberacochea (approved):
Thanks @SandyRogers

PR #338, opened 2026-04-29
Import assemblies from the filesystem flow.

mberacochea:
This PR:
- Adds an import assemblies from the filesystem flow
- Minor refactoring to prevent duplicated code with this new flow and existing ones
- If the coverage_json has missing keys it will throw an exception now (in the old and new flows -- https://github.com/EBI-Metagenomics/emgapi-v2/compare/feature/import-assemblies-from-the-filesystem?expand=1#diff-a439466dad30558d1831c5733684563f7b58a77e0d49bb9fbdd6deddf3ea8371L106)

This flow takes a miassembler results directory and registers the assemblies in the DB; it will do minimal sanity checks on the provided folders. Users are responsible for making sure the results are consistent, and that the accessions in the results match the study that is provided.

This is meant as a solution for those studies that we assembled in other infra (we have some that we assembled in AWS); or those that require customizing loads of the parameters of miassembler to be tweaked... or even as a convenience way to register the assemblies in the DB.

The results are not copied over to NFS for archiving; that is the users’ responsibility. If people feel strongly about this, I can add a task to do this too.

---

#### Checklist


analyses/models.py
```diff
@@ -588,6 +589,31 @@ def dir_with_miassembler_suffix(self):
             / Path(assembler.version)
         )
 
+    def update_coverage_metadata_from_report(
+        self, coverage_report: Mapping[str, Any]
```
SandyRogers: Presumably more specifically 
```suggestion
        self, coverage_report: Mapping[str, str | float | None]
```
?

analyses/models.py
```diff
@@ -588,6 +589,31 @@ def dir_with_miassembler_suffix(self):
             / Path(assembler.version)
         )
 
+    def update_coverage_metadata_from_report(
```
SandyRogers, line 592: Would be tempted to generalise this method into the `InferredMetadataMixin`, so like 
```python
class InferredMetadataMixin:
...
def populate_metadata_from_json(self, json: Mapping[str, Any], required_keys: list[str], allowed_keys: list[str])
```

.talismanrc
```diff
@@ -225,6 +225,12 @@ fileignoreconfig:
 - filename: workflows/tests/test_flows_utils.py
   checksum: 7044790c62db033597a0c9391751a32f2db1f88d5ba5a953b6c607d97ff7ed6d
 
+- filename: analyses/models.py
+  allowed_patterns: [key]
```
SandyRogers: try and keep this file vaguely alphabetically ordered for easier scanning! not crucial, obviously.

workflows/flows/assemble_study_tasks/miassembler_reports.py
```diff
@@ -0,0 +1,81 @@
...
+                dtype=str,
+            )
+        )
+    except pd.errors.EmptyDataError:
+        return QcFailedRunsReport.validate(
+            pd.DataFrame(columns=QcFailedRunsReport.to_schema().columns)
+        )
+
+
+def read_coverage_report(path: Path) -> Any:
```
SandyRogers: could we call this (and other similar method) `load_*_report` instead of `read_*_report` ? just because of the word-clash with genomic reads, i.e. a report about read coverage of the assembly contigs

workflows/flows/assemble_study_tasks/miassembler_reports.py
```diff
@@ -0,0 +1,81 @@
...
+import pandera.pandas as pa
+from pandera.typing import DataFrame, Series
+
+
+class AssembledRunsReport(pa.DataFrameModel):
+    """
+    Schema for miassembler rows describing successfully assembled runs.
+    """
+
+    run_accession: Series[str] = pa.Field(nullable=False)
```
SandyRogers: should we have regex on the accession formats? or are you purposefully allowing them to be non-INSDC?

workflows/flows/upload_assembly.py
SandyRogers, line 1: not driven by this PR, but this should probably switch to the django_flow etc in case of long-running flow (like exponentially backed off retries of the upload)

workflows/flows/shared/study_tasks.py
```diff
@@ -0,0 +1,27 @@
+from prefect import get_run_logger
+
+from analyses.models import Study
+import ena.models
+from workflows.ena_utils.ena_api_requests import get_study_from_ena
+from workflows.prefect_utils.flows_utils import django_db_task as task
+
+
+@task
```
SandyRogers, line 9: Hmm, does this need to be inside a task boundary?

workflows/flows/shared/run_tasks.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+@task
+def get_or_import_mgnify_run(
+    run_accession: str,
+    mgnify_study_id: int,
+    read_run_limit: int,
+) -> int:
+    """
+    Get an MGnify Run by accession, fetching the study's ENA read runs first if needed.
```
SandyRogers, line 15: I'm not completely against this, but it has a hint of backwardness to me – having a run trigger a study-level fetch suggests things have not been ordered well in a parent flow (which should set up its Study first) ?

workflows/flows/import_assemblies_from_filesystem_flow.py
```diff
@@ -0,0 +1,433 @@
...
+    """
+    logger = get_run_logger()
+    assembled_runs_csv = Path(nextflow_outdir) / "assembled_runs.csv"
+    report = read_assembled_runs_report(assembled_runs_csv)
+    logger.info(f"Read {len(report)} assembled run records")
+    return report
+
+
+@task(name="Read QC failed runs report")
+def read_qc_failed_runs(nextflow_outdir: str | Path) -> DataFrame[QcFailedRunsReport]:
```
SandyRogers: similar to elsewhere, I'd prefer `load_` as the verb in the naming here to avoid confusion with genomic `read`s

workflows/flows/import_assemblies_from_filesystem_flow.py
```diff
@@ -0,0 +1,433 @@
...
+        logger.info(
+            f"{'Created' if created else 'Updated'} assembly {assembly.id} for run {run_accession}"
+        )
+        return assembly.id
+
+
+@task(name="Get study assembly records from ENA")
+def get_study_assembly_records_from_ena(
+    study_accession: str,
+    mgnify_study_id: int,
```
SandyRogers: why ID rather than MGYS accession?

workflows/data_io_utils/miassembler_utils.py
```diff
@@ -0,0 +1,25 @@
+from pathlib import Path
+
+from workflows.data_io_utils.filenames import accession_prefix_separated_dir_path
+
+
+MIASSEMBLER_ACCESSION_PREFIX_LENGTH = 7
+
+
+def miassembler_run_output_dir(
```
SandyRogers, line 9: Does this maybe just belong in `workflows.data_io_utils` somewhere? E.g. `workflows.data_io_utils.miassembly_utils` ? Not important.

SandyRogers (commented):
Thanks @mberacochea ! Various minor/pedantic comments, happy with the overall plan though.

.talismanrc
mberacochea: ah, I wasn't sure what was the convention.. I just kept adding them at the end. I shall sort them!

workflows/flows/assemble_study_tasks/miassembler_reports.py
mberacochea: Ah, yeah. I forgot about that, I was thinking that we need some more defensive bits in this flow.

workflows/flows/shared/study_tasks.py
mberacochea: My take is that with the django tasks have the db connection cleaning bit for free.

workflows/flows/assemble_study_tasks/miassembler_reports.py
mberacochea: yeah, good point.. I haven't thought about clashing with reads. I'll rename things

workflows/flows/import_assemblies_from_filesystem_flow.py
mberacochea: Yup

workflows/flows/import_assemblies_from_filesystem_flow.py
mberacochea: no particular reason, I'll use the study accession for consistency

workflows/data_io_utils/miassembler_utils.py
mberacochea: Maybe, I put that there to begin with. 🤔

workflows/flows/shared/run_tasks.py
mberacochea: I agree, I took this from the assemble flow (but there is different). I'll simplify this, pull the study then the study runs... easier to follow and less side effects.

analyses/models.py
mberacochea: yup

analyses/models.py
mberacochea: gave this a stab, there is a method there which we should use in other places too

workflows/flows/assemble_study_tasks/miassembler_reports.py
mberacochea: sorted

workflows/flows/assemble_study_tasks/miassembler_reports.py
mberacochea: panderas regex check :done:

workflows/flows/shared/run_tasks.py
mberacochea: I've simplified this... but left this method (I'll remove it later on)

workflows/data_io_utils/miassembler_utils.py
mberacochea: Moved to data_io_utils... it looks better there I think *thanks*!

workflows/flows/shared/study_tasks.py
SandyRogers: Fair

analyses/base_models/base_models.py
```diff
@@ -285,6 +286,35 @@ def __getitem__(self, item):
...
+        required_keys: list[str],
+        allowed_keys: list[str],
+    ) -> None:
+        """
+        Copy selected values from a JSON-like mapping into ``metadata``.
+        """
+        if not isinstance(json_data, Mapping):
+            raise ValueError("JSON data must be a mapping")
+
+        unexpected_keys = sorted(set(json_data) - set(allowed_keys))
```
SandyRogers: I probably led you down a bad path with the `allowed_keys` naming. I think I really meant `optional_keys`, and we would import the union of `required_keys + optional_keys`, but only raise exception if any required are not present.
(So that the JSON source could have other ignorable keys, and we're just specifying which ones we NEED and which ones we ALSO WANT). But whatever, this is completely fine too if it works for this use case.

workflows/ena_utils/ena_accession_matching.py
```diff
@@ -18,6 +18,11 @@
 )
 ENA_ASSEMBLY_ACCESSION_GLOB: str = f"[{_INSDC_CENTRE_PREFIXES}]RZ[0-9]*"
 
+INSDC_RUN_ACCESSION_REGEX: re.Pattern = re.compile(
+    f"([{_INSDC_CENTRE_PREFIXES}]RR[0-9]{{6,}})"
+)
+INSDC_RUN_ACCESSION_GLOB: str = f"[{_INSDC_CENTRE_PREFIXES}]RR[0-9]*"
+
```
SandyRogers, line 25: great

SandyRogers: BTW, I think this probably means we don't support co-assembly style `;`-delimited accession in this. That is fine I think, but maybe add a TODO or a line to the import flow to make that clear?

SandyRogers (approved):
:shipit: (FYI future musings below)

analyses/base_models/base_models.py
mberacochea: yeah, it was late and I was tired. This implementation hsa the potential to break things. I'll swap this one

workflows/ena_utils/ena_accession_matching.py
mberacochea: Indeed

mberacochea: Sorted

analyses/base_models/base_models.py
mberacochea: sorted!

PR #339, opened 2026-04-30
Add support to TPA in the import miassembler from fs

mberacochea:
Add support to TPA in the import miassembler from fs

---

#### Checklist

### DIFF
--- a/workflows/flows/import_assemblies_from_filesystem_flow.py
+++ b/workflows/flows/import_assemblies_from_filesystem_flow.py
@@ -135,7 +135,7 @@ def validate_completed_assembly_files(
 @task(name="Import completed assembly")
 def import_completed_assembly(
     reads_mgnify_study_id: int,
-    assembly_submission_mgnify_study_id: int | None,
+    tpa_mgnify_study_accession: str | None,
     run_id: int,
     nextflow_outdir: Path,
     record: dict[str, str],
@@ -148,7 +148,7 @@ def import_completed_assembly(
 
     :param reads_mgnify_study_id: MGnify Study ID for the reads used to produce
         the assembly.
-    :param assembly_submission_mgnify_study_id: MGnify Study ID for the separate
+    :param tpa_mgnify_study_accession: MGnify Study accession for the separate
         assembly submission study when importing TPA assemblies. Pass ``None`` for
         non-TPA imports, where the assembly belongs to the reads study.
     :param run_id: MGnify Run ID for the assembled read run.
@@ -164,8 +164,8 @@ def import_completed_assembly(
     logger = get_run_logger()
     reads_mgnify_study = analyses.models.Study.objects.get(id=reads_mgnify_study_id)
     assembly_submission_mgnify_study = (
-        analyses.models.Study.objects.get(id=assembly_submission_mgnify_study_id)
-        if assembly_submission_mgnify_study_id
+        analyses.models.Study.objects.get(accession=tpa_mgnify_study_accession)
+        if tpa_mgnify_study_accession is not None
         else None
     )
     ena_study = (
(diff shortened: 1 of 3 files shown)


workflows/flows/import_assemblies_from_filesystem_flow.py
```diff
@@ -135,18 +134,46 @@ def validate_completed_assembly_files(
 
 @task(name="Import completed assembly")
 def import_completed_assembly(
-    mgnify_study_id: int,
+    reads_mgnify_study_id: int,
+    assembly_submission_mgnify_study_id: int | None,
```
SandyRogers: In general I prefer passing accessions rather than IDs (just because they are more immediately obvious what they are in logs etc). We should probably add this to the Style Guide in the README...

SandyRogers: `assembly_submission_mgnify_study_id`... maybe naming like `tpa_mgnify_study_accession` would be more appropriate here since you want None if the assembly study is just the reads study... which means this is only ever supposed to be set for TPAs?

workflows/flows/import_assemblies_from_filesystem_flow.py
```diff
@@ -135,18 +134,46 @@ def validate_completed_assembly_files(
...
+    :param overwrite_existing_assemblies: Allow updating an existing assembly row for
+        the same run/sample pair instead of failing.
+    :return: The created or updated Assembly ID.
     """
     logger = get_run_logger()
-    mgnify_study = analyses.models.Study.objects.get(id=mgnify_study_id)
+    reads_mgnify_study = analyses.models.Study.objects.get(id=reads_mgnify_study_id)
+    assembly_submission_mgnify_study = (
+        analyses.models.Study.objects.get(id=assembly_submission_mgnify_study_id)
+        if assembly_submission_mgnify_study_id is not None
```
SandyRogers:
```suggestion
        if assembly_submission_mgnify_study_id
```

since there is no 0 study ID (but also using accessions would be nicer for this check too)

workflows/flows/import_assemblies_from_filesystem_flow.py
```diff
@@ -155,22 +182,30 @@ def import_completed_assembly(
...
+                f"Assembly for run {run_accession} already exists in the database"
+            )
 
         assembly.assembler = assembler
-        assembly.reads_study = mgnify_study
-        assembly.ena_study = mgnify_study.ena_study
+        assembly.reads_study = reads_mgnify_study
+        assembly.assembly_study = assembly_submission_mgnify_study
+        # TPA imports follow the submission study.
+        assembly.ena_study = ena_study
```
SandyRogers, line 210: this is a good point... this is currently correct but logically probably incorrect. I just had a discussion with @arajkovic about removing these `ena_study` objects anyway as they are now essentially tech debt IMO. Just mentioning here for awareness

SandyRogers (approved):
I think it looks sensible, thanks @mberacochea

workflows/flows/import_assemblies_from_filesystem_flow.py
mberacochea: Good point, I'm using ids just because it is "faster" to get. I'll swap that and a line in the style guide.

workflows/flows/import_assemblies_from_filesystem_flow.py
mberacochea: yup

workflows/flows/import_assemblies_from_filesystem_flow.py
mberacochea: Alright, I'll add a TODO comment here

workflows/flows/import_assemblies_from_filesystem_flow.py
mberacochea: yeah, that is a good pint. Using tpa_ is makes it easier to follow

PR #341, opened 2026-05-01
Housekeeping flow -> fix eggnogmapper duplicated headers

mberacochea:
This flow is probably an overkill, but with the robots is very easy to write so instead of a one-of-script I decided to create this mini-monster.

The idea is that it is going to check for duplicated headers in eggnogmapper tsv files, if there any it will correct them. It is quite defensive, it will create a backup copy of the files first and it has a dry_run option and a dont_sync_to_ftp; to prevent messing up production results.

---

#### Checklist

### DIFF
--- a/workflows/flows/housekeeping/fix_eggnog_tsv_headers.py
+++ b/workflows/flows/housekeeping/fix_eggnog_tsv_headers.py
@@ -34,6 +34,7 @@ class EggnogHeaderFixCandidate(TypedDict, total=False):
     external_path: str
     has_duplicate_header: bool
     backup_path: str
+    synced_to_ftp: bool
 
 
 def _analysis_eggnog_annotation_downloads(analysis: Analysis):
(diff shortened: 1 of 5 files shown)


workflows/flows/housekeeping/fix_eggnog_tsv_headers.py
```diff
@@ -0,0 +1,319 @@
...
+) -> list[EggnogHeaderFixCandidate]:
+    """Collect files with duplicated EggNOG headers and publish a review artifact.
+
+    :param analysis_accessions: Optional accession filter.
+    """
+    logger = get_run_logger()
+    candidates: list[EggnogHeaderFixCandidate] = []
+    for analysis in _eggnog_analyses(analysis_accessions):
+        candidates.extend(_build_candidate_rows(analysis))
+
```
arajkovic:
```suggestion
    candidates: list[EggnogHeaderFixCandidate] = [
        _build_candidate_rows(analysis)
        for analysis in _eggnog_analyses(analysis_accessions)
    ]

```

workflows/flows/housekeeping/fix_eggnog_tsv_headers.py
```diff
@@ -0,0 +1,319 @@
...
+    if not sync_to_ftp:
+        logger.info("sync_to_ftp=False - Flow complete after NFS repair.")
+        return repaired
+
+    if not repaired:
+        logger.info("No repaired analyses to sync to FTP.")
+        return []
+
+    resync_eggnog_results_to_ftp(repaired)
```
arajkovic: Maybe it should be reported which ones have been synced, since those are already returned from this task.

arajkovic (approved):
Left a couple suggestions. The prefect side of things seems alright to me.

workflows/flows/housekeeping/fix_eggnog_tsv_headers.py
mberacochea: good idea, I've changed that.

PR #351, opened 2026-05-06
Set the concurrency_limit to 5 for asa/virify/map flows

mberacochea:
Limits the number of concurrent ASA / VIRify and MAP batch flows to 5. So at max there could be 50 assemblies being ASAed + 50 VIRified + 50 MAPed; this cap should help with the CPUs/MEM quota in the cluster.

---

#### Checklist


SandyRogers (approved):
Thanks @mberacochea   
Maybe also add to the donco yaml so we can apply locally as a loose form of validation? 

Idea for later: run the prefect cli on this file in GHA to validate it too.

PR #360, opened 2026-05-14
Refactored the fs sync/copy results code -- Fixed missing assembly analyses NFS result_dir

mberacochea:
The idea is to simplify the code that copy the results files to NFS/FTP (and sync the status in the DB). This code uses the different pipelines PipelineSchema (used to validate results) to define where files should be copied over (to create the final results folder structure). This is useful for assembly analysis where the results are scattered across 3 different pipelines.

This can be further refactored but I think it is a good starting point.

---

#### Checklist


workflows/flows/housekeeping/fix_eggnog_tsv_headers.py
```diff
@@ -70,18 +70,8 @@ def _eggnog_analyses(analysis_accessions: list[str] | None = None):
 
 
 def analysis_results_root(analysis: Analysis) -> Path:
```
SandyRogers: this seems like a redundant method now?

workflows/tests/test_fix_eggnog_tsv_headers_flow.py
```diff
@@ -20,11 +20,14 @@ def _write_gzipped_tsv(path: Path, lines: list[str]) -> None:
...
-def test_build_candidate_rows_supports_batch_root_results_dir():
+ASSEMBLY_EXTERNAL_RESULTS_DIR = "PRJEB105/PRJEB105754/ERZ28775/ERZ28775516/V6/assembly"
+
+
+def test_build_candidate_rows_supports_mirrored_nfs_results_dir():
     analysis = Mock(
         accession="MGYA00000001",
-        results_dir="/tmp/batch-root",
-        external_results_dir="analyses/MGYA00000001",
+        results_dir=f"/tmp/nfs/{ASSEMBLY_EXTERNAL_RESULTS_DIR}",
```
SandyRogers: should this dir be from a pytest tmpdir fixture perhaps?

workflows/flows/analyse_study_tasks/shared/copy_v6_pipeline_results.py
```diff
@@ -57,15 +81,15 @@ def copy_v6_pipeline_results(analysis_accession: str, timeout: int = 14400):
...
 
-    target_root = (
+    destination_root = (
         EMG_CONFIG.slurm.private_results_dir
         if analysis.is_private
         else EMG_CONFIG.slurm.ftp_results_dir
     )
 
-    target = f"{target_root}/{accession_prefix_separated_dir_path(study.first_accession, -3)}/{accession_prefix_separated_dir_path(analysis.assembly_or_run.first_accession, -3)}/{analysis.pipeline_version}/{experiment_type_label}"
+    destination = f"{destination_root}/{accession_prefix_separated_dir_path(study.first_accession, -3)}/{accession_prefix_separated_dir_path(analysis.assembly_or_run.first_accession, -3)}/{analysis.pipeline_version}/{experiment_type_label}"
```
SandyRogers: Minor improvement here would be to join this with pathlib to avoid any potential `//`

workflows/flows/analyse_study_tasks/shared/copy_v6_pipeline_results.py
```diff
@@ -183,134 +207,96 @@ def copy_v6_study_summaries(
...
     logger.info(
-        f"Study {study} now has results at {study.external_results_dir} in {target_root}"
+        f"Study {study} now has results at {study.external_results_dir} in {destination_root}"
     )
 
 
 @task(
-    name="Copy Assembly Batch Results",
-    task_run_name="Copy Assembly Batch Results for batch {batch_id}",
+    description="Copy an Assembly Batch Results",
```
SandyRogers:
```suggestion
    description="Copy an Assembly Batch's Results",
```

workflows/flows/analysis/assembly/flows/sync_assembly_batch_results.py
```diff
@@ -0,0 +1,153 @@
...
+        analysis = analyses[result.analysis_id]
+        analysis.status[Analysis.AnalysisStates.ANALYSIS_ANNOTATIONS_IMPORTED] = (
+            result.success
+        )
+        if result.success:
+            analysis.status[Analysis.AnalysisStates.ANALYSIS_QC_FAILED] = False
+            analysis.status[
+                Analysis.AnalysisStates.ANALYSIS_POST_SANITY_CHECK_FAILED
+            ] = False
+            analysis.status[Analysis.AnalysisStates.ANALYSIS_BLOCKED] = False
```
SandyRogers: 2 possible issues here?
1. we have a status setter method already that unsets other statuses (like ANALYSIS_BLOCKED) when setting a new incoherent status (like ..IMPORTED). Maybe can just use that / extend it to work in batch if needed?
2. If I am following right... we set ANALYSIS_ANNOTATIONS_IMPORTED only on the basis that the data copy worked. Is there a possibility that the data -> db ingestion fails, but copy succeeds, in which case this status is incorrect? I worry we could have an edge case where the DB connection drops (for example) whilst ingesting taxonomies, but we can't tell that from the statuses.

workflows/flows/analysis/assembly/flows/sync_assembly_batch_results.py
```diff
@@ -0,0 +1,153 @@
...
+        analysis = analyses[result.analysis_id]
+        destination = Path(result.destination_folder)
+        analysis.external_results_dir = destination.relative_to(destination_root)
+        analysis.study.external_results_dir = destination.parent.relative_to(
+            destination_root
+        )
+        studies_to_update[analysis.study_id] = analysis.study
+
+    Analysis.objects.bulk_update(analyses.values(), ["external_results_dir"])
+    Study.objects.bulk_update(studies_to_update.values(), ["external_results_dir"])
```
SandyRogers: I can't quite think of the scale that we expect here... is bulk_update a premature optimisation? The only real reason to NOT use it is because it doesn't call django hooks; we do have a few hooks already and are considering some more search indexes etc.

SandyRogers (commented):
Thanks @mberacochea 

I left a few comments, they may be not real issues but some are worth checking quickly if possible

workflows/flows/analysis/assembly/flows/sync_assembly_batch_results.py
mberacochea: yeah, we could do individual updates to be honest. Each batch is max 10 assemblies... I'll use that

workflows/flows/analysis/assembly/flows/sync_assembly_batch_results.py
mberacochea: 1. Yes, me and the robots forgot about that
2. :hmm: I'm missing a filter, only those that reached COMPLETED should be considered!. Good catch.

workflows/flows/housekeeping/fix_eggnog_tsv_headers.py
mberacochea: Yes, _gone_

workflows/flows/analyse_study_tasks/shared/copy_v6_pipeline_results.py
mberacochea:
```
    destination = (
        Path(destination_root)
        / accession_prefix_separated_dir_path(study.first_accession, -3)
        / accession_prefix_separated_dir_path(
            analysis.assembly_or_run.first_accession, -3
        )
        / analysis.pipeline_version
        / experiment_type_label
    )
```

workflows/flows/analysis/assembly/flows/sync_assembly_batch_results.py
```diff
@@ -0,0 +1,174 @@
...
+    external_results_root = (
+        EMG_CONFIG.slurm.private_results_dir
+        if mgnify_study.is_private
+        else EMG_CONFIG.slurm.ftp_results_dir
+    )
+    # Store the study roots up front. Individual Analysis rows will later point deeper
+    # into the ENA-derived tree at
+    # {ena_study_prefix}/{ena_assembly_prefix}/{pipeline_version}/assembly.
+    mgnify_study.external_results_dir = str(study_prefix)
+    mgnify_study.save(update_fields=["external_results_dir"])
```
arajkovic: Maybe move the model save after it has been successfully copied.

workflows/flows/analyse_study_tasks/shared/copy_v6_pipeline_results.py
```diff
@@ -183,134 +215,88 @@ def copy_v6_study_summaries(
...
+    destination_root: Path,
     timeout: int = 14400,
-):
+) -> BatchCopyResult:
     """
-    Copy results for a single analysis from the batch workspace to external results.
+    Copy results for a single assembly analysis from the batch workspace to external results.
 
-    FIXME: this method and copy_v6_pipeline_results are _very_ similar. This duplication is a problem
+    This only considers the batch analysis what that are COMPLETED (as pre the status in the batch analysis).
```
arajkovic:
```suggestion
    This only considers the batch analysis that are COMPLETED (as pre the status in the batch analysis).
```

workflows/flows/analysis/assembly/flows/sync_assembly_batch_results.py
```diff
@@ -0,0 +1,174 @@
...
+        analysis = analyses[result.analysis_id]
+        analysis.external_results_dir = str(
+            study_prefix
+            / accession_prefix_separated_dir_path(
+                analysis.assembly_or_run.first_accession, -3
+            )
+            / analysis.pipeline_version
+            / Analysis.ExperimentTypes.ASSEMBLY.label.lower()
+        )
+        analysis.save()
```
arajkovic, line 147: Batch update would be more efficient here, I think.

workflows/flows/analysis/assembly/flows/sync_assembly_batch_results.py
```diff
@@ -0,0 +1,174 @@
...
+        analysis = analyses[result.analysis_id]
+        analysis.results_dir = str(
+            study_results_dir
+            / accession_prefix_separated_dir_path(
+                analysis.assembly_or_run.first_accession, -3
+            )
+            / analysis.pipeline_version
+            / Analysis.ExperimentTypes.ASSEMBLY.label.lower()
+        )
+        analysis.save()
```
arajkovic, line 175: Same here, batch update.

arajkovic (approved):

workflows/flows/analysis/assembly/flows/sync_assembly_batch_results.py
mberacochea: yes, Sandy's argument is that the batches don't trigger signals. And each batch is at most 10 analysis so I think this is fine.

workflows/flows/analysis/assembly/flows/sync_assembly_batch_results.py
mberacochea: Same as before.

workflows/flows/analysis/assembly/flows/sync_assembly_batch_results.py
mberacochea: Good point!

workflows/flows/analysis/assembly/flows/sync_assembly_batch_results.py
arajkovic: Makes sense. Should have read his comments more carefully 😄

workflows/flows/analysis/assembly/flows/sync_assembly_batch_results.py
mberacochea: done

PR #361, opened 2026-05-15
Fixed issue causing malformed genome catalogue download file links

MGS-sails:
This PR:
Ftp links on Genome catalogues were being malformed due to issues with urljoin() in the resolve_url methhod of Download file schemas.
When a relative path has a preceding slash, url join treats it as an absolute path and this affects how the download links get built. 
This issue was resolved by ensuring that values coming from genome_catalogue.results_dir have no preceding slashes. 

This PR also refactors the resolve_url method, by introducing a new helper: https://github.com/EBI-Metagenomics/emgapi-v2/blob/c25c6f2d4941fead5c68c7bbcf221d00d3a97014/genomes/schemas/url_utils.py

the resolve_url methods of both MGnifyGenomeDownloadFile and MGnifyGenomeCatalogueDownloadFile both make use of that to reduce code repetition
*

---

#### Checklist

### DIFF
--- a/genomes/schemas/MGnifyGenomeCatalogueDownloadFile.py
+++ b/genomes/schemas/MGnifyGenomeCatalogueDownloadFile.py
@@ -2,18 +2,15 @@ from __future__ import annotations
 
 import logging
 from typing import Union
-from urllib.parse import urljoin
 
-from django.conf import settings
 from ninja import Field, Schema
 from typing_extensions import Annotated
 
 from analyses.base_models.with_downloads_models import DownloadFile
 from genomes import models as genome_models
-from workflows.data_io_utils.filenames import trailing_slash_ensured_dir
+from genomes.schemas.url_utils import build_transfer_service_public_url
 
 logger = logging.getLogger(__name__)
-EMG_CONFIG = settings.EMG_CONFIG
 
 
 class MGnifyGenomeCatalogueDownloadFile(Schema, DownloadFile):
@@ -45,7 +42,4 @@ class MGnifyGenomeCatalogueDownloadFile(Schema, DownloadFile):
             # Without a results directory, we cannot form a URL
             return None
 
-        return urljoin(
-            EMG_CONFIG.service_urls.transfer_services_url_root,
-            urljoin(trailing_slash_ensured_dir(catalogue.result_directory), obj.path),
-        )
+        return build_transfer_service_public_url(catalogue.result_directory, obj.path)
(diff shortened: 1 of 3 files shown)


genomes/schemas/url_utils.py
```diff
@@ -0,0 +1,28 @@
...
+
+from pathlib import Path
+from urllib.parse import urljoin
+
+from django.conf import settings
+
+from workflows.data_io_utils.filenames import trailing_slash_ensured_dir
+
+
+def build_transfer_service_public_url(
```
mberacochea: I would rename `build_public_url` or `get_public_url`.

genomes/schemas/url_utils.py
```diff
@@ -0,0 +1,28 @@
...
+    result_directory: str | Path | None, file_path: str | Path | None
+) -> str | None:
+    """
+    Build a public download URL for a file that lives under a results directory
+    exposed by the Transfer Service.
```
mberacochea:
```suggestion
    Build a public download URL for a file that lives under a results directory
    exposed in EBI FTP server.
```
"Transfer Services" is the team, I would be explicit

genomes/schemas/MGnifyGenomeCatalogueDownloadFile.py
```diff
@@ -45,7 +42,4 @@ def resolve_url(obj: "MGnifyGenomeCatalogueDownloadFile"):
             # Without a results directory, we cannot form a URL
```
mberacochea, line 42: I would log an error here :sentry-to-rescue:

genomes/schemas/url_utils.py
```diff
@@ -0,0 +1,28 @@
...
+    - Ensures the joined path is relative (no leading slash) so the base path is preserved.
+    """
+    transfer_service_root_url = (
+        settings.EMG_CONFIG.service_urls.transfer_services_url_root
+    )
```
mberacochea: I would use `settings.EMG_CONFIG.service_urls.transfer_services_url_root` directly

mberacochea (approved):
A few pedantic comments. 

Can you add an example of how the URLs are transformed please?

SandyRogers (approved):
Thanks @MGS-sails .

This suggests the results_dirs for some catalogues in the DB are absolute paths instead of relative paths.. so I think we also need to modify the data (can do through admin, no need for code)

PR #362, opened 2026-05-18
De-duplication commands for Studies and Runs.

mberacochea:
2 flows to de-duplicate studies and runs.

🔴 -> Also patched update_or_create_by_accession to always check the ena_accessions to prevent duplicates in the future... as much as possible.

Codex helped me write this one.

---

#### Checklist


workflows/flows/housekeeping/deduplicate_runs.py
```diff
@@ -0,0 +1,253 @@
...
+            results.append(result)
+            logger.info(
+                f"Finished group {index}/{len(duplicated_runs)} "
+                f"with canonical run {result.canonical_run}"
+            )
+        except Exception as exc:
+            logger.exception(
+                f"Failed to deduplicate run group {[run.id for run in runs]}",
+                exc_info=exc,
+            )
```
arajkovic: I think this exception should append a `GroupMergeResult` to `results` with the error field populated.

analyses/base_models/base_models.py
```diff
@@ -125,9 +125,7 @@ def update_or_create_by_accession(
         # Two-step process needed (first get, then update/create).
         # This is because __overlap cannot be used with default django update_or_create.
         try:
-            obj = self.get_queryset().get(
-                ena_accessions__overlap=known_accessions, **kwargs
-            )
+            obj = self.get_queryset().get(ena_accessions__overlap=known_accessions)
```
arajkovic, line 133: Maybe adjust the docstring so it clearly states that `kwargs` is not used for filtering anymore.

workflows/flows/housekeeping/deduplicate_runs.py
mberacochea: yeah, good idea

analyses/base_models/base_models.py
mberacochea: Yes, good point.

SandyRogers (changes requested):
Thanks @mberacochea 

(__This may be opinion__) but I think this would be more appropriately handled by a Django management command (ideally) rather than a flow. Mostly because this is more of a non-schema-changing migration of bad data. I would approach more like:
- django command(s) to the de-dupe (we already have `merge_mgys_duplicates`)
- add a helper to e.g. `ENADerivedModel` to add a db-level constraint on `ena_accessions` field and migration to apply it once the command has been run, therefore command/flow "never" needed again

If you do strongly prefer flows for this though then can this PR also remove the existing implementation of `merge_mgys_duplicates` ?

Happy to chat about approach if easier!

mberacochea:
> Thanks @mberacochea
> 
> (**This may be opinion**) but I think this would be more appropriately handled by a Django management command (ideally) rather than a flow. Mostly because this is more of a non-schema-changing migration of bad data. I would approach more like:
> 
>     * django command(s) to the de-dupe (we already have `merge_mgys_duplicates`)
> 

A command is fine, I'm thinking in flow-terms quite a lot lately - 🔨:nail: situation. A command is probably even better - the only thing that our flows/tasks give us is the cleaning of old_connections (not sure how relevant it will be here though).

I didn't use merge_mgys_duplicates as the basis, because it wasn't covering all the cases I though we needed.

>     * add a helper to e.g. `ENADerivedModel` to add a db-level constraint on `ena_accessions` field and migration to apply it once the command has been run, therefore command/flow "never" needed again
> 

Yes. I added the change to `update_or_create_by_accession` just to avoid inserting any more duplicated by accident.

> 
> If you do strongly prefer flows for this though then can this PR also remove the existing implementation of `merge_mgys_duplicates` ?
> 
> Happy to chat about approach if easier!

analyses/base_models/base_models.py
```diff
@@ -104,14 +104,19 @@ def update_or_create_by_accession(
...
+        to identify duplicate records.
+        `kwargs` are only applied when a new object is created. After update or creation,
+        `ena_accessions` is set to include the union of provided and existing accessions.
+
+        :param known_accessions: Accessions to match against ``ena_accessions`` and preserve on the object.
         :param defaults: Fields to update.
         :param create_defaults: Fields to set if the object is created.
         :param include_update_defaults_in_create_defaults: If true, the defaults dict will be merged with create_defaults for creation.
-        :param kwargs: other matchers
+        :param kwargs: Extra fields to apply only when creating a new object.
```
MGS-sails, line 119: Should kwargs be mentioned here seeing as it's been removed from params ?

analyses/management/commands/merge_runs_duplicates.py
```diff
@@ -0,0 +1,347 @@
...
+        all_accessions = set(canonical.ena_accessions or [])
+        for duplicate in duplicates:
+            all_accessions.update(duplicate.ena_accessions or [])
+
+        merged_accessions = sorted(all_accessions)
+        if sorted(canonical.ena_accessions or []) != merged_accessions:
+            canonical.ena_accessions = merged_accessions
+            canonical.save(update_fields=["ena_accessions"])
+
+    def rewire_run_relations(self, canonical_run: Run, duplicates: list[Run]) -> None:
```
MGS-sails, line 149: Do studies and other relations like sample also need to be rewired to the canonical run ?

analyses/management/commands/merge_runs_duplicates.py
```diff
@@ -0,0 +1,347 @@
...
+                        assembly_id=assembly_id,
+                        run_id=canonical_run.id,
+                    )
+                    for assembly_id in duplicate_assembly_ids
+                ],
+                ignore_conflicts=True,
+            )
+            assembly_run_through.objects.filter(run_id=duplicate.id).delete()
+
+    def run_is_fully_detached(self, run: Run) -> bool:
```
MGS-sails, line 181: Might be worth having a test for a run that is fully detached and can't be deleted

analyses/management/commands/merge_runs_duplicates.py
```diff
@@ -0,0 +1,347 @@
...
+        """
+        Check whether a Run has any remaining direct relations that block deletion.
+
+        :param run: Run row being considered for deletion after rewiring.
+        :return: True if no direct tracked relations remain attached.
+        """
+        return not any(
+            [
+                run.analyses.exists(),
```
MGS-sails, line 190: should related studies and samples also be checked here ?

MGS-sails (commented):
Thanks @mberacochea I've left some comments

analyses/base_models/base_models.py
mberacochea: Is still used, not to `get` the record but used downstream in the method

analyses/management/commands/merge_runs_duplicates.py
mberacochea: yes, I dropped the ball a bit with the runs. I'll add those checks

mberacochea: Well, I checked the data and there are no duplicated samples. Runs have a sample_id field, so as soon as those duplicated are gone that is it.

analyses/management/commands/merge_runs_duplicates.py
mberacochea: The studies and samples are already re-wired in merge_mgys_duplicates.py

analyses/management/commands/merge_runs_duplicates.py
mberacochea: As far as I can see in the DB, the samples are fine (no dups). Duplicated runs are linked to the same sample, so deleting the runs should be fine

analyses/management/commands/merge_ena_study_duplicates.py
```diff
@@ -0,0 +1,417 @@
...
+        """
+        Load duplicate ENA Study groups by overlapping accessions.
+
+        :param accessions: Optional ENA accessions used to limit which groups are returned.
+        :return: Duplicate groups whose accession aliases place them in the same connected component.
+        """
+        groups: list[list[ena_models.Study]] = []
+        requested = set(accessions or [])
+
+        with connection.cursor() as cursor:
```
SandyRogers, line 70: BTW, any particular reason to do this with sql rather than models?

analyses/management/commands/merge_ena_study_duplicates.py
```diff
@@ -0,0 +1,417 @@
...
+        return not any(
+            [
+                Study.objects.filter(ena_study=ena_study).exists(),
+                Sample.objects.filter(ena_study=ena_study).exists(),
+                Run.objects.filter(ena_study=ena_study).exists(),
+                Assembly.objects.filter(ena_study=ena_study).exists(),
+                Analysis.objects.filter(ena_study=ena_study).exists(),
+                ena_models.Sample.objects.filter(study=ena_study).exists(),
+            ]
+        )
```
SandyRogers, line 257: I was thinking we could also use
```
from django.db.models.deletion import Collector
collector = Collector(using="default")
collector.collect([ena_study]
```

to get the list of relations... but I think your version is more explicit tbh 👍

SandyRogers (approved):
thanks @mberacochea !

analyses/management/commands/merge_ena_study_duplicates.py
mberacochea: I couldn't figure out how to run a query like this with the ORM, and this command should be :nuked: soon

analyses/management/commands/merge_ena_study_duplicates.py
mberacochea: yeah

PR #374, opened 2026-06-04
Migrate import_primer_identification from a flow to a command

mberacochea:
This kind of data correction commands should live in mgm commands instead of flows.

---

#### Checklist


MGS-sails (approved):

workflows/management/commands/import_primer_identification_for_imported_analyses.py
```diff
@@ -0,0 +1,109 @@
...
+        )
+
+    def handle(self, *args, **options):
+        """
+        Import available primer-identification outputs for eligible amplicon analyses.
+
+        :param args: Positional command arguments, unused.
+        :param options: Parsed command options.
+        """
+        max_count = options["max_count"]
```
SandyRogers, line 72: Would suggest adding something here to verify that the command is running in a place that has access to the fs needed for this... i.e. we are running HPC not on K8s.

Could add a helper or subclass `Command` or more simply

```suggestion
        assert Path(EMG_CONFIG.slurm.default_workdir).exists()
        max_count = options["max_count"]
```

SandyRogers (commented):
Thanks @mberacochea just one suggestion otherwise happy once tests are passing

workflows/management/commands/import_primer_identification_for_imported_analyses.py
mberacochea: Ah.. good point. Let's go with this patch now as I would like to nuke this command after running it.

PR #378, opened 2026-06-11
Fix retries on prefect (and fix donco dev slurm cluster)

SandyRogers:
This PR:
* adds a `key` for every flow-run-input when we do a `suspend_flow_run(wait_for_input...)`
    * there is a new tiny util to make a key that CHANGES on every flow-run attempt, i.e. it changes after a retry. This is counter-intuitive from the prefect docs, which suggest we would NOT want this behaviour, and it usually picks a random key that is fixed for the flow-run's lifecycle to avoid duplicate suspensions. There seems to be an edge case when the input is used in a suspend, with retries, but it wasn't immediately obvious how to fix this in prefect itself so I solved it locally, by ensuring the flow-run-input is newly asked on every retry. Better logic might involve explicitly conditionally suspending, but that needs detailed per-flow logic.
    * this fixes a long-standing bug where pressing "Retry" in the prefect UI for a flow with a suspend/wait-for-input gave an error like `409, 'A flow run input with this key already exists.`
* fixes the `donco` dev slurm cluster, which was unable to actually run any flow-runs.
  * some config was needed for the slurm-node to use postgres, and to activate the right python env, and some missing env vars etc.
  * I mostly did this because I needed to be able to reproduce the 409 bug to figure it out, but as a side note we should be making better use of this local env for e2e purposes.

---

#### Checklist


workflows/prefect_utils/input_helpers.py
```diff
@@ -0,0 +1,5 @@
+from prefect.runtime import flow_run
+
+
+def ask_every_time_suspend_for_input_key():
+    return f"suspend-{flow_run.id}-attempt-{flow_run.run_count}"
```
mberacochea, line 10:
```suggestion
def ask_every_time_suspend_for_input_key():
    """
    Generate a flow-run input key that changes on each retry attempt.
    This avoids 409 conflicts when retrying flows that suspend and wait for input,
    by ensuring a new flow-run input is created for every attempt.
    """
    return f"suspend-{flow_run.id}-attempt-{flow_run.run_count}"
```

mberacochea (approved):
Thanks @SandyRogers .. I just left a tiny comment.

PR #379, opened 2026-06-12
adds rss feed of latest analysed studies

SandyRogers:
This PR:
* adds an RSS feed of the latest public studies
* mostly because _I wanted this feature_ ... but PDBe and EBISearch also provide RSS feeds so it is with precedent 

---

#### Checklist

### DIFF
--- a/emgapiv2/feeds.py
+++ b/emgapiv2/feeds.py
@@ -1,0 +1,41 @@
+from urllib.parse import urljoin
+
+from django.conf import settings
+from django.contrib.syndication.views import Feed
+
+from analyses.models import Study
+
+
+class LatestStudiesFeed(Feed):
+    title = "MGnify Latest Studies"
+    link = "/studies/"
+    description = "Latest studies analysed by MGnify pipelines"
+
+    def items(self):
+        return Study.public_objects.order_by("-updated_at")[
+            : settings.EMG_CONFIG.distribution.latest_studies_feed_count
+        ]
+
+    def item_title(self, item: Study):
+        return f"{item.accession}: {item.title}"
+
+    def item_description(self, item: Study):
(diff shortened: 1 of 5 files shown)


emgapiv2/feeds.py
```diff
@@ -0,0 +1,41 @@
...
+        return item.metadata.get("center_name", "No submission centre available.")
+
+    def item_pubdate(self, item: Study):
+        return item.created_at
+
+    def item_updateddate(self, item: Study):
+        return item.updated_at
+
+    def item_categories(self, item: Study):
+        return [item.biome.biome_name]
```
mberacochea, line 41: It would be nice to also include the analysis type here, and have that as a filter.

mberacochea (approved):
Nice, RSS are a great thing.

emgapiv2/feeds.py
SandyRogers: I thought about exactly that yeh - and then started thinking about how to handle mixed-type studies, and what about ones where we had run e.g. Amplicon pipeline, so only fetched Amplicon reads, but then it is kind of incorrect to call it an "amplicon study" if the INSDC study actually also has WGS reads that we "don't know about". And that all seemed a bit complicated to know what is "correct" so I bravely did nothing.

SandyRogers: I will give this a more general thought though – because generally representing the analysis state of a study would be useful in various places (like EBI Search index). It has always annoyed me that you can't search for all "V5 Amplicon Soil" studies there. So even with the above caveats/complications, this would be a nice thing to do universally in future.

mberacochea: Yes, I think the flow should be... tag a study as "amplicon" from the analyses. We would need to retrofit this, but once it is done the tagging should be easy.

PR #380, opened 2026-06-18
Better library source/strategy control

SandyRogers:
This PR:
* adds more options for overriding library source & strategy of ENA run metadata
  * the existing `ENALibraryStrategyPolicy` now also propagates to the `Run.experiment_type` field in the DB, rather than only controlling what gets put on pipeline samplesheets. I.e., it is persisted.
* a new `ENALibrarySourcePolicy` is added, similar to the strategy policy. 
  * It has a default value which means if a run's scientific-name metadata is like "*metagenome", we ignore the fact that the library source is `GENOMIC` or `OTHER` – two very common incorrect metadata cases.
  * like the library strategy policy, the library source policy can be chosen when starting/retrying a prefect analysis flow run. It too is persisted so it helps get "correct" (i.e., curated) `Run.experiment_type` values as well as just making the right runs appear on sample sheets.

Together these should by default handle cases like an amplicon run of these reads:

```tsv
run_accession	library_layout	library_strategy	library_source	tax_id	scientific_name
SRR22284878	PAIRED	AMPLICON	GENOMIC	408172	marine metagenome
SRR22284880	PAIRED	AMPLICON	GENOMIC	408172	marine metagenome
```

There is also a management command to try and migrate older Runs and Analyses to this updated logic, e.g. there are 20K Runs with UNKNOWN in the MGnify DB, and 4K of those just have the wrong `library_source` which would be corrected here.

---

2 unrelated bug fixes.

---

#### Checklist

### DIFF
--- a/analyses/management/commands/determine_unknown_run_experiment_types.py
+++ b/analyses/management/commands/determine_unknown_run_experiment_types.py
@@ -155,7 +155,9 @@ class Command(BaseCommand):
             "analysis_accessions": ",".join(
                 run.analyses.order_by("accession").values_list("accession", flat=True)
             ),
-            "library_strategy": metadata.get(Run.CommonMetadataKeys.LIBRARY_STRATEGY, ""),
+            "library_strategy": metadata.get(
+                Run.CommonMetadataKeys.LIBRARY_STRATEGY, ""
+            ),
             "library_source": metadata.get(Run.CommonMetadataKeys.LIBRARY_SOURCE) or "",
             "scientific_name": metadata.get(Run.CommonMetadataKeys.SCIENTIFIC_NAME)
             or "",
--- a/analyses/management/commands/determine_unknown_run_experiment_types.py
+++ b/analyses/management/commands/determine_unknown_run_experiment_types.py
@@ -155,8 +155,7 @@ class Command(BaseCommand):
             "analysis_accessions": ",".join(
                 run.analyses.order_by("accession").values_list("accession", flat=True)
             ),
-            "library_strategy": metadata.get(Run.CommonMetadataKeys.LIBRARY_STRATEGY)
-            or "",
+            "library_strategy": metadata.get(Run.CommonMetadataKeys.LIBRARY_STRATEGY, ""),
             "library_source": metadata.get(Run.CommonMetadataKeys.LIBRARY_SOURCE) or "",
             "scientific_name": metadata.get(Run.CommonMetadataKeys.SCIENTIFIC_NAME)
             or "",
--- a/analyses/management/commands/determine_unknown_run_experiment_types.py
+++ b/analyses/management/commands/determine_unknown_run_experiment_types.py
@@ -1,0 +1,176 @@
+from __future__ import annotations
+
+import csv
+import logging
+from pathlib import Path
+
+from django.core.management.base import BaseCommand
+
+from analyses.models import Analysis, Run
+from workflows.ena_utils.ena_policies import (
+    ENALibrarySourcePolicy,
+    ENALibraryStrategyPolicy,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = (
+        "Determine UNKNOWN run experiment types from analyses, assemblies, or ENA "
+        "metadata, and write a TSV report of changed and unresolved runs."
+    )
+
+    report_headers = [
+        "run_accession",
+        "study_accession",
+        "analysis_accessions",
+        "library_strategy",
+        "library_source",
+        "scientific_name",
+        "previous_experiment_type",
+        "new_experiment_type",
+        "status",
+        "reason",
+    ]
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--output-tsv",
+            default="unknown_run_experiment_type_report.tsv",
+            help="Path to the TSV report to write.",
+        )
+
+    def handle(self, *args, **options):
+        output_tsv = Path(options["output_tsv"])
+        report_rows = []
+        changed_count = 0
+        unresolved_count = 0
+
+        runs = (
+            Run.objects.filter(experiment_type=Run.ExperimentTypes.UNKNOWN)
+            .select_related("study", "study__ena_study")
+            .prefetch_related("analyses", "assemblies")
+        )
+        total_count = runs.count()
+        logger.info(f"Checking {total_count} UNKNOWN run experiment types")
+
+        for run in runs:
+            previous_experiment_type = run.experiment_type
+            reason = self.determine_experiment_type(run)
+            run.refresh_from_db()
+
+            if run.experiment_type != previous_experiment_type:
+                changed_count += 1
+                self.inherit_changed_run_experiment_type(run)
+                logger.info(
+                    f"Changed run {run.first_accession} from {previous_experiment_type} to {run.experiment_type} using {reason}"
+                )
+                report_rows.append(
+                    self.report_row(run, previous_experiment_type, "changed", reason)
+                )
+            elif run.experiment_type == Run.ExperimentTypes.UNKNOWN:
+                unresolved_count += 1
+                logger.warning(
+                    "Could not determine experiment type for run %s",
+                    run.first_accession,
+                )
+                report_rows.append(
+                    self.report_row(run, previous_experiment_type, "unresolved", reason)
+                )
+
+        self.write_report(output_tsv, report_rows)
+        logger.info(
+            "Finished checking UNKNOWN run experiment types: %s changed, %s unresolved, %s reported",
+            changed_count,
+            unresolved_count,
+            len(report_rows),
+        )
+        self.stdout.write(
+            self.style.SUCCESS(
+                f"Checked {total_count} UNKNOWN runs; changed {changed_count}; "
+                f"unresolved {unresolved_count}; wrote {output_tsv}"
+            )
+        )
+
+    def determine_experiment_type(self, run: Run) -> str:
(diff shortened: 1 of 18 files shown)


workflows/ena_utils/ena_api_requests.py
```diff
@@ -262,6 +275,10 @@ def _make_samples_and_run(
     run.set_experiment_type_by_metadata(
```
mberacochea, line 276: I think this is a good idea, and I like the implementation.

One thing, when we override `self.experiment_type = expected_experiment_type`... should we keep track of that with a bool?. It may come in handy in the future, as there will be a discrepancy with ENA

workflows/flows/assemble_study.py
```diff
@@ -165,6 +173,24 @@ class AssembleStudyInput(RunInput):
...
 
+    if (
+        assemble_study_input.library_strategy_policy
+        != ENALibraryStrategyPolicy.ONLY_IF_CORRECT_IN_ENA
+        or assemble_study_input.library_source_policy
+        != ENALibrarySourcePolicy.OVERRIDE_GENOMIC_IF_METAGENOMIC_SCIENTIFIC_NAME
+    ):
+        read_runs = get_study_readruns_from_ena(
+            ena_study.accession,
+            limit=5000,
```
mberacochea: Not for this PR, but we need to account for the few edge cases where there are more than 5k runs.

workflows/flows/assemble_study.py
```diff
@@ -165,6 +173,24 @@ class AssembleStudyInput(RunInput):
     validate_and_set_webin_owner(ena_study, assemble_study_input.webin_owner)
     mgnify_study.refresh_from_db()
 
+    if (
```
mberacochea, line 176: I just realised, do we only pull the read_runs if we change the policy?

analyses/management/commands/determine_unknown_run_experiment_types.py
```diff
@@ -0,0 +1,176 @@
...
+        return {
+            "run_accession": run.first_accession or "",
+            "study_accession": (
+                run.study.ena_study.accession if run.study.ena_study else ""
+            ),
+            "analysis_accessions": ",".join(
+                run.analyses.order_by("accession").values_list("accession", flat=True)
+            ),
+            "library_strategy": metadata.get(Run.CommonMetadataKeys.LIBRARY_STRATEGY)
+            or "",
```
mberacochea:
```suggestion
            "library_strategy": metadata.get(Run.CommonMetadataKeys.LIBRARY_STRATEGY, ""),
```

mberacochea (approved):
Looks good to me, I left some questions in there.

:shipit:

workflows/ena_utils/ena_api_requests.py
SandyRogers: I pondered this. In the `metadata` JSON itself, we have the whole `InferredMetadataMixin` setup to let us store and use `inferred_*` keys whilst preserving the original (ENA-matching) keys. My argument is that `experiment_type` is a MGnify-only field, it doesn't actually map to any single thing in ENA, so it is *always* inferred by some logic somewhere (whether automatic when created here, or automatic in a migration, or by a team member's logic in admin panel).

workflows/flows/assemble_study.py
SandyRogers: Nah it is also called on line 103. It is a bit ugly, but in all these flows we call it once in the first iteration of the flow-run, to see what we have. Then if the admin-user changes policy we call it again to pick up potentially more runs.

workflows/flows/assemble_study.py
SandyRogers: Done here anyway, as good a place as any.

arajkovic (approved):

analyses/management/commands/determine_unknown_run_experiment_types.py
```diff
@@ -0,0 +1,177 @@
...
+            ena_library_strategy=metadata.get(Run.CommonMetadataKeys.LIBRARY_STRATEGY)
+            or "",
+            ena_library_source=metadata.get(Run.CommonMetadataKeys.LIBRARY_SOURCE)
+            or "",
+            ena_scientific_name=metadata.get(Run.CommonMetadataKeys.SCIENTIFIC_NAME)
+            or "",
+            library_strategy_policy=ENALibraryStrategyPolicy.ONLY_IF_CORRECT_IN_ENA,
+            library_source_policy=ENALibrarySourcePolicy.OVERRIDE_GENOMIC_IF_METAGENOMIC_SCIENTIFIC_NAME,
+        )
+        return "ena_metadata"
```
chrisAta, line 138: do we not need a `run.save()` here? probably misunderstanding how this works

chrisAta (approved):
looks good to me!! just one question to make sure I understood

analyses/management/commands/determine_unknown_run_experiment_types.py
SandyRogers: Good q! We don't in this case because `set_experiment_type_by_metadata` saves itself IFF the experiment type changes when run. This is also why there is a `run.refresh_from_db()` straight after the invocation of this function.

analyses/management/commands/determine_unknown_run_experiment_types.py
```diff
@@ -0,0 +1,177 @@
...
+
+        # if the run has an assembly, assume it was metagenomic
+        if run.assemblies.exists():
+            run.experiment_type = Run.ExperimentTypes.METAGENOMIC
+            run.save()
+            return "assembly_attached"
+
+        # otherwise use the default handling of ENA policies...
+        # this will still leave some UNKNOWN because we don't know the context in which runs were being fectched,
+        # e.g. if we were in an "amplicon analysis flow" we'd have assumed/ unknown things are probably amplicon
```
Ge94, line 126: Would it make sense then to have a similar approach for amplicon and set any UNKNOWN to AMPLICON (or METAGENOMIC?), maybe during the final import or at a stage where we surely know the analysis worked?

analyses/management/commands/determine_unknown_run_experiment_types.py
```diff
@@ -0,0 +1,177 @@
...
+            analysis.inherit_experiment_type()
+
+    def report_row(
+        self, run: Run, previous_experiment_type: str, status: str, reason: str
+    ) -> dict[str, str]:
+        metadata = run.metadata_preferring_inferred
+        return {
+            "run_accession": run.first_accession or "",
+            "study_accession": (
+                run.study.ena_study.accession if run.study.ena_study else ""
```
Ge94, line 153: Are there realistically cases where we don't have any run or study? Or is this just being overly cautious?

Just wondering whether a "" would lead to errors down the line and it would be better to stop the execution here...

analyses/tests/test_models.py
```diff
@@ -28,6 +34,62 @@ def create_analysis(is_private=False):
...
+    assert run.experiment_type == Run.ExperimentTypes.AMPLICON
+
+    # Case from issue: Metagenome scientific name + GENOMIC + WGS
+    run.set_experiment_type_by_metadata(
+        "WGS", "GENOMIC", ena_scientific_name="marine metagenome"
+    )
+    assert run.experiment_type == Run.ExperimentTypes.METAGENOMIC
+
+    # Unknown case: GENOMIC + AMPLICON (without metagenome scientific name)
+    run.set_experiment_type_by_metadata(
```
Ge94, line 62: So if I understand correctly, this run wouldn't be allowed into an amplicon flow, right?

workflows/ena_utils/ena_policies.py
```diff
@@ -0,0 +1,29 @@
+from emgapiv2.enum_utils import FutureStrEnum
+
+ALLOWED_LIBRARY_SOURCE: list = ["METAGENOMIC", "METATRANSCRIPTOMIC"]
+COMMONLY_INCORRECT_LIBRARY_SOURCE: list = ["GENOMIC", "OTHER"]
```
Ge94, line 4: Is "UNKNOWN" part of these?

Ge94 (approved):
Looks good! I just left silly questions to make sure I understood 🥹

analyses/management/commands/determine_unknown_run_experiment_types.py
SandyRogers: This code comment may not be clearly written enough – but yes what you suggest is indeed already there in the analysis flows (i.e. UNKNOWN typically becomes AMPLICON in an amplicon analysis flow). What this comment is saying is that in the context of this migration command ONLY, we may not know anything about the run and have no context (like an analysis flow) to go on.

analyses/management/commands/determine_unknown_run_experiment_types.py
SandyRogers: I think probably overly cautious yes – although this is just for printing a report so that `""` shouldn't matter 👍

analyses/tests/test_models.py
SandyRogers: Correct yep

workflows/ena_utils/ena_policies.py
SandyRogers: Nope. `"UNKNOWN"` is one of our MGnify terms (for `experiment_type`, which is our collapsed combination of `library_source` and `library_strategy` with modifications using `scientific_name` and manual curation). So things can be `UNKNOWN` for various reasons, one of which could be that in ENA `library_source` is `"OTHER"` with insufficient/incoherent other values to correct it. Full list of possible library source values is here: https://ena-docs.readthedocs.io/en/latest/submit/reads/webin-cli.html#permitted-values-for-library-source

PR #381, opened 2026-06-19
MGYS for ERP/PRJ etc accessions

SandyRogers:
This PR:
* adds API endpoint `studies/insdc/{accession}` which takes an ERP/SRP/PRJ etc and returns an MGYS accession
  * this is so we can support redirecting INSDC study/project accessions to MGnify project accessions especially on the website
  * I considered adding redirects, Link headers or Content Location headers, to do redirecting from the API... but they all felt like they were blurring the web-client/API divide too much
* adds GIN indexes to all the models with an `ena_accessions` field. These were missing because the existing index is just a btree which doesn't help the `__contains` lookup.  It is around 100x cheaper to do the lookup that this endpoint needs with the GIN index.

---

#### Checklist


arajkovic (approved):
:shipit:

vagkaratzas (approved):
Seems sound

mberacochea (approved):
It looks good to me. I guess the client will do the redirect (?).

We should also generate a dump with MGYS-ERP/PJR (maybe)

SandyRogers:
> It looks good to me. I guess the client will do the redirect (?).
> 
> We should also generate a dump with MGYS-ERP/PJR (maybe)

Yep on both counts!

PR #382, opened 2026-06-19
ASA contaminant reference genome setting

mberacochea:
This allows users to set a contaminant reference genome per study for the assembly analysis pipeline decontamination module.

There is no way to set this per sample/assembly ATM.

Codex helped with the code.

---

#### Checklist


workflows/tests/test_nextflow_utils.py
SandyRogers, line 1: not sure that this file needs to be updated at all? it is essentially a dummy implementation

SandyRogers (commented):
Great, thanks Martin. Happy with the idea of doing on the ASA batch object. 
Main suggestion is to consider integrating it with the existing (MIAssembler-focussed) `HostReferenceGenome` implementation in `host_metadata_to_reference_genome.py`

Specific ideas:
- Use `KNOWN_REFERENCE_GENOMES` to make `contaminant_reference` a django choice field with that list as the choices... if you use the `DjangoChoicesCompatibleStrEnum` thing it would probably mean you get both a nice dropdown in django admin, and a nice dropdown in prefect suspend UI if/when that step is added. And validation that the contaminant reference exists.
- auto-populate the `contaminant_reference` using host-tax-id and host-scientific-name kind of metadata using existing `get_reference_genome_for_host` 


This all assumes that ASA looks for contaminant genomes in the same place as MIAssembler – I didn't actually check this.

Ignore this if you want, but if easy for codex to do it would avoid another divergence I think.

ochkalova (approved):
Thanks Martin!
I don't completely understand the changes, but what I understand looks good 👍🏻

workflows/tests/test_nextflow_utils.py
mberacochea: probably.. I just renamed that just-in-case.

mberacochea:
> `HostReferenceGenome`

Yes, I went with no validation to deploy this one faster (and because I forgot about the miassembler ref selection). But I think we need to be more flexible and support ad-hoc genomes too (or make it easy to add more genomes). For example, Sonya needs some weird ones.
I'll merge and deploy this one, to unblock Sonya and then will implement the known genomes thingy.

PR #384, opened 2026-06-22
Tidy up a change to handle assemblies that not linked to any runs

mberacochea:
This should prevent cases when the assemblies are not linked to any runs. Which is correct as ENA allows this.

This came up with some of Sonya's @ochkalova studies. Specially PRJEB88595, @ochkalova can you check if the primary metagenomes in this study are not linked to any runs please?

---

#### Checklist

### DIFF
--- a/workflows/ena_utils/ena_api_requests.py
+++ b/workflows/ena_utils/ena_api_requests.py
@@ -595,8 +595,9 @@ def get_study_assemblies_from_ena(accession: str, limit: int = 10) -> list[str]:
         else:
             run = mgnify_sample.runs.first()  # TODO: coassemblies? replicates?
 
+        assembly_accession = assembly_data[_.ANALYSIS_ACCESSION]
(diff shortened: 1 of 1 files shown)


SandyRogers (approved):
Thanks @mberacochea . Looks like a sensible change.

I think in the case mentioned, the assembly does have run refs (but co-assembly style):
```tsv
analysis_accession	analysis_type	run_accession	sample_accession	assembly_type
ERZ26273096	SEQUENCE_ASSEMBLY	ERR13917198;ERR13925930	SAMEA118006056	primary metagenome
```

so in principle we could add another case here to determine a reads study based on the run-refs of the assemblies... but overall I think it potentially fills MGnify with unanalysed Studies and is fine to leave all that layer of connection to be explored in ENA

ochkalova (approved):
LGTM 👍🏻

PR #385, opened 2026-06-24
EMG-9948 create metadata json for dwc summaries script

ochkalova:
This pull request refactors the Darwin Core Ready (DwC-R) summary generation workflow to remove its dependency on live ENA API queries for run and sample metadata. Instead, it builds the required per-run metadata directly from the ingested Study, Run, and Sample objects and writes this to a local JSON file, which is then used by the summary generator. Test code has also been updated to reflect this change, removing unnecessary API mocks and ensuring the test data is set up locally.

**DwC-R summary generation task:**

* Added `build_run_metadata_json` in `dwcr_generator.py` to construct a per-run metadata mapping from Study, Run, and Sample objects, outputting all necessary fields and filling missing values with "NA". This eliminates the need for live ENA API queries during summary generation.
* Updated `generate_dwc_ready_summary_for_pipeline_run` to use the locally-generated metadata JSON file, passing it to the summary generator instead of relying on ENA API lookups.

**Testing and test data setup:**

* Removed all ENA API mocking for run and sample metadata from `test_analysis_amplicon_study_flow.py` and `test_dwcr_generator.py`, as the workflow no longer requires live API calls for these data. 
* Added `_ensure_flow_like_amplicon_objects_for_dwcr` helper to `test_dwcr_generator.py` to ensure test studies, runs, and samples have the minimum required metadata for DwC-R summary generation, mimicking the state after ENA ingestion.

---

#### Checklist


workflows/tests/test_dwcr_generator.py
```diff
@@ -159,8 +88,15 @@ def test_dwcr_merge(
...
+    summaries_folder = amplicon_folder / "summaries"
+    summaries_folder.mkdir(exist_ok=True, parents=True)
+    logging.info(summaries_folder)
+    logging.info(summaries_folder.exists())
+
     # create fake dwcr files for testing
-    with open(amplicon_folder / "abc123_DADA2_SILVA_16S-V3-V4_dwcready.csv", "w") as fw:
+    with open(
+        summaries_folder / "abc123_DADA2_SILVA_16S-V3-V4_dwcready.csv", "w"
+    ) as fw:
```
ochkalova, line 98: Maybe I'm wrong but I had to make this change because merging script expects tsv files to merge in `summaries/` folder, not in `amplicon_v6/` folder. Without this change, this test fails.

workflows/flows/analyse_study_tasks/shared/dwcr_generator.py
```diff
@@ -35,6 +37,51 @@
...
+            "SampleID": sample_obj.first_accession,
+            "StudyID": study.first_accession,
+            "decimalLatitude": _get_meta(sample_obj, "lat"),
+            "decimalLongitude": _get_meta(sample_obj, "lon"),
+            "collectionDate": _get_meta(sample_obj, "collection_date"),
+            "depth": _get_meta(sample_obj, "depth"),
+            "temperature": _get_meta(sample_obj, "temperature"),
+            "salinity": _get_meta(sample_obj, "salinity"),
+            "country": _get_meta(sample_obj, "country"),
+            "InstitutionCode": _get_meta(sample_obj, "center_name"),
```
SandyRogers: There are constants for these strings in e.g. `Run.CommonMetadataKeys.LAT` etc. Two reasons to use those instead of strings:
1. it makes code editor's code intelligence help you by spotting typos / missing keys
2. those constants are derived from a snapshot of ENA's API, so if one of these fields is renamed/removed there, it means we can find the usages more easily to update in future

SandyRogers (commented):
Thanks @ochkalova !

I think the need for a temporary JSON file to transport the metadata from this repo's responsibility to the mgnify-pipelines-toolkit responsibility suggests we should in future refactor the DwC generator into this codebase directly. But that can be a future TODO!

ochkalova:
> I think the need for a temporary JSON file to transport the metadata from this repo's responsibility to the mgnify-pipelines-toolkit responsibility suggests we should in future refactor the DwC generator into this codebase directly. But that can be a future TODO!

Yes, exactly. Martin is doing this 👍🏻 
This solution is temporary

workflows/flows/analyse_study_tasks/shared/dwcr_generator.py
ochkalova: Makes sense... Let me add this

ochkalova: [refactor: use metadata fields constants in dwcr_generator](https://github.com/EBI-Metagenomics/emgapi-v2/pull/385/commits/e8345fa65205cae6063a038bdeeb1867a7c736f4)

mberacochea (approved):
LGTM. Thanks for fixing this one @ochkalova

PR #390, opened 2026-07-20
Move sourmash gather into emgapi-v2 with Django Tasks worker

MGS-sails:
This PR:
*
This PR brings the sourmash gather flow in-house within emgapi-v2, replacing the old split-service pattern with an API + Django Tasks worker model.

## What changed  
Added a new /genomes-search/gather/ flow in emgapi-v2
Introduced Django Tasks + django-tasks-db as the async execution backend
Added GenomeSearchIndex, SourmashSearchJob, and SourmashSearchJobItem models
Added a db_worker-based app-worker service for running sourmash searches
Implemented sourmash task execution and result/status tracking in-app
Split worker dependencies from the main web app runtime to avoid the prefect / sourmash dependency conflict
Seeded local dev data so a searchable sourmash catalogue is available in make-dev-data
Added admin visibility for the new search/index/job models
Added test coverage for gather submission, status, results, and queue failure cases.

The sourmash search flow depends on the same catalogue/version state as the main API and database, so treating it as a separate microservice adds operational complexity without giving much real separation. This moves search closer to the source of truth, keeps metadata enrichment in one place, and gives us a reusable async pattern that could also work well for Branchwater and cobs later.

Requires running the Django Tasks worker (app-worker) for async gather execution
make-dev-data now seeds a local searchable sourmash catalogue for end-to-end testing

Added/updated unit tests for the gather API flow
Verified local end-to-end flow with API submission, queued task execution, and worker processing

---

#### Checklist


emgapiv2/api/genome_search_gather.py
```diff
@@ -0,0 +1,471 @@
...
+
+
+def _parse_uuid(raw_id: str) -> Optional[uuid.UUID]:
+    try:
+        return uuid.UUID(str(raw_id))
+    except (TypeError, ValueError):
+        return None
+
+
+def _get_active_sourmash_indexes(mag_catalogues: set[str]) -> List[GenomeSearchIndex]:
```
mberacochea, line 146: This one should be cached as this changes very rarely

MGS-sails: @mberacochea I added a caching to the function. It caches the ids of search indexes

emgapiv2/api/tests/test_genome_search_gather.py
```diff
@@ -0,0 +1,267 @@
...
+from analyses.models import Biome
+from genomes.models import (
+    GenomeCatalogue,
+    GenomeSearchIndex,
+    SourmashSearchJob,
+    SourmashSearchJobItem,
+)
+
+
+def _make_catalogue() -> GenomeCatalogue:
```
SandyRogers: if we use an @fixture here instead, then this can inherit the `top_level_biomes` fixture and be a bit simpler

docker-compose.yaml
```diff
@@ -125,16 +134,44 @@ services:
...
+      - DATABASE_URL=postgres://postgres:postgres@app-database:5432/emg
+      - PREFECT_API_URL=http://prefect-server:4200/api
+      - PREFECT_UI_URL=http://localhost:4200
+      - EMG_SOURMASH__QUERIES_PATH=/app/tools/sourmash-local/data/queries
+      - EMG_SOURMASH__RESULTS_PATH=/app/tools/sourmash-local/data/results
+      - EMG_SOURMASH__INDEX_ROOT=/app/tools/sourmash-local/data/signatures
+      - EMG_WEBIN__JWT_SECRET_KEY=dummy-key
+      - EMG_WEBIN__AUTH_ENDPOINT=http://httpbin/status/200
+      - EMG_WEBIN__TOKEN_ENDPOINT=http://httpbin/status/200
+      - EMG_WEBIN__ACCOUNT_DETAILS_ENDPOINT=http://httpbin/status/200
```
SandyRogers: does the app-worker actually need these set btw?

Dockerfile
```diff
@@ -20,12 +20,37 @@ RUN apt -y update && \
         python3-wheel
 
 WORKDIR /app
-COPY requirements.txt .
+COPY requirements-common.txt requirements.txt requirements-sourmash-worker.txt .
```
SandyRogers: I would simplify this build a bit, if possible, and use an identical set of requirements for the api container and the sourmash-worker container. The only downside, I think, is that one more package (sourmash) is installed in the API and not currently used by it. But that is a pretty minor downside I think? That way on prod both the api and worker could use identical docker image, which is quite handy for tidiness.

requirements-common.txt
SandyRogers, line 1: See comment above on the Dockerfile – I think splitting the requirements just to avoid installing sourmash package on the api is probably unneeded.

sourmash_search_architecture_plan.md
SandyRogers, line 1: This and the design plan above are handy docs to see and keep – but better moved to a dev journal on confluence than merged.

tools/sourmash-local/smoke_test.sh
SandyRogers, line 1: Nice. If this is intended for ongoing dev helpfulness (I think it is and should be?) can we have a taskfile entry for it e.g. `task debug_sourmash_smoke_test` or similar.

docker-compose.yaml
```diff
@@ -125,16 +134,44 @@ services:
...
       - "8000:8000"
     networks: [emg]
 
+  app-worker:
+    # Django Tasks worker for sourmash searches
+    build:
+      context: .
+      target: sourmash_worker
+    container_name: app-worker
+    profiles: ["app", "all"]
```
SandyRogers: I wonder if this should only be included when `all` is explicitly picked, to keep the default compose stack a tad lighter?
```suggestion
    profiles: ["all"]
```

tools/sourmash-local/docker-compose.yml
SandyRogers, line 1: Is this one redundant now? Just a leftover from the migration from apiv1 right?

tools/sourmash-local/README.md
SandyRogers, line 1: I guess this is also leftover right as we aren't using celery now?

emgapiv2/config.py
```diff
@@ -289,6 +289,15 @@ class ServiceURLsConfig(BaseModel):
     genome_search_proxy: str = "https://cobs-genome-search-01.mgnify.org/search"
 
 
+class SourmashConfig(BaseModel):
```
SandyRogers, line 292: should the scaling factor of 1000 default be in here too?

emgapiv2/api/genome_search_gather.py
```diff
@@ -0,0 +1,503 @@
...
+class SourmashGatherStatusOut(Schema):
+    data: SourmashGatherStatusData
+
+
+def _api_path(path_suffix: str) -> str:
+    return f"/{settings.BASE_URL}{path_suffix.lstrip('/')}"
+
+
+def _absolute_api_url(request, path_suffix: str) -> str:
+    return request.build_absolute_uri(_api_path(path_suffix))
```
SandyRogers: I think better to use the EMG_CONFIG's Service URLs config section for this ?

emgapiv2/api/genome_search_gather.py
```diff
@@ -0,0 +1,503 @@
...
+        return []
+    if hasattr(files, "getlist"):
+        return [_coerce_uploaded_file(f) for f in files.getlist("file_uploaded")]
+
+    uploaded = files.get("file_uploaded")
+    if uploaded is None:
+        return []
+    if isinstance(uploaded, list):
+        return [_coerce_uploaded_file(f) for f in uploaded]
+    return [_coerce_uploaded_file(uploaded)]
```
SandyRogers: It isn't too obvious to me what is happening here / why we need this coercion etc. Is it still needed? If so, could we have a comment here to explain what is happening?

genomes/models/sourmash_search_job.py
```diff
@@ -0,0 +1,159 @@
...
+import uuid
+
+from django.db import models
+
+from analyses.base_models.base_models import TimeStampedModel
+
+from .genome_search_index import GenomeSearchIndex
+
+
+class SourmashSearchJob(TimeStampedModel):
```
SandyRogers: I am probably missing something having not used django-tasks properly before. But, why do we need these models, vs just letting django-tasks' default model handle the storage of the search job? I thought that was the default way to use it, just like:

```
from django_tasks import task

@task()
def sourmash_search(catalogue, query_file):
    return sourmash_gather(stuff, ...)

and in the api:

job = sourmash_search.enqueue('human_gut', Path(...))
return {'job': job.id, 'results': '/results/{job.id}'}
```

Is there some guidance as to when one should add these custom models to represent the task jobs?

SandyRogers (commented):
Thanks @MGS-sails ! Very glad we're bringing this all together. A few comments/questions in line but bigger one is the question I left about task models on `sourmash_search_job.py`.

Dockerfile
MGS-sails: I agree a single shared image would be cleaner, however, there is a Prefect/sourmash dependency conflict, when prefect and sourmash are put into thesame reqs file. and then a sourmash import-time segfault in the worker build, so I think it’s safer to keep the worker requirements isolated for now.

genomes/sourmash.py
```diff
@@ -0,0 +1,184 @@
...
+    name_map_path: str = "",
+) -> dict[str, Any]:
+    from sourmash.commands import SaveSignaturesToLocation
+    from sourmash.search import format_bp, gather_databases
+    from sourmash.sourmash_args import (
+        FileOutputCSV,
+        get_moltype,
+        load_dbs_and_sigs,
+        load_query_signature,
+    )
```
mberacochea, line 88: Why are these imports here?

tools/sourmash-local/docker-compose.yml
```diff
@@ -0,0 +1,45 @@
+---
+name: sourmash-local
+
+services:
+  redis:
+    image: redis:7
+    container_name: sourmash-local-redis
```
mberacochea: I guess this was part of the old implementation and can be removed

docker-compose.yaml
```diff
@@ -125,16 +134,44 @@ services:
...
+    command: "python manage.py db_worker --skip-checks"
```
mberacochea, line 154: Why the `skip--checks`?

sourmash_django_tasks_detailed_design.md
mberacochea, line 1: I agree with Sandy these are more suitable for Confluence

genomes/models/sourmash_search_job.py
mberacochea: django-tasks-db backend has the models, and even a basic admin

MGS-sails: @SandyRogers I've changed the tasks to work on a one job per request basis. 
I've removed the custom Sourmash Job and Sourmash Job item models. 

@mberacochea we're currently making use of the models and showing the db task table records on the admin panel

emgapiv2/api/genome_search_gather.py
MGS-sails: This code existed to ensure all data coming from the request is in a valid format. 
Although as was discussed recently, the web client is the only user of the sourmash service, so we have more control over what data get's through as such i will remove this code

docker-compose.yaml
MGS-sails: This was a debugging step at the time. It's now removed

genomes/sourmash.py
MGS-sails: These are kept here to prevent other files that rely on other functions in the sourmash file, from having to import these files 

There are other files that import functions from genomes/sourmash.py, but do not require the heavy sourmash dependencies. 
Keeping those dependencies inside this function, rather than at the top of the file, prevents those other files from having unnecessary dependencies.

PR #392, opened 2026-07-22
Update-able genome catalogues and import flow

SandyRogers:
This PR:
* Adds more tasks to the genome import flow, specifically ones related to moving data between file systems and preparing sourmash indexes etc. Essentially things that used to be in the Jenkins pipeline. See https://github.com/EBI-Metagenomics/emgapi-v2/compare/EMG-9938?expand=1#diff-dd2935cc445f63ad05dfc96f37ccf2811627457d6f272824d195c726d68e5669 (some of this is a bit TODO still because of the overlap with changes happening to the sourmash. Also the COBS stuff will be much improved when moves to k8s, it looks quite ugly as a port from a jenkins job....)
* Significantly changes the Genome/Catalogue etc models, to separate a Genome instance (the MGYG accessioned sequence really) from its annotations, which are actually not the accessioned object... they are catalogue-version dependent and change over time. So, these are now reflected in the `CatalogueGenome` model. CatalogueGenome also is like a thru model for a Genome being in multiple Catalogues. Why? Two reasons... firstly, because in a single rw database model, without snapshot releases, there was no way to update a catalogue without some downtime or inconsistent data since the process is too long for a db transaction. This model lets e.g. marine-v1 and marine-v2 – which share many genomes – coexist. This is done by introducing a model for the CatalogueSeries (i.e., the versionless "marine") which can have both published and draft catalogues in it – only one is published at a time though. A new auth layer is added, whereby a user calling the API with the config's EMG webin account token can see draft catalogues. A few route/query-param changes were needed to, which mean an auth'd user on the website can preview the catalogue before publishing it (via the admin panel) which also unpublishes any previous version. The other reason for the m2m genome<>catalogue relationship is if an MGYG really does exist in multiple catalogues, e.g. because of blurry biome boundaries or if one is a superset of another. There are migrations, tests, and dev-data helpers for all of this. Many of the file additions are just dev fixtures for this.

---

#### Checklist


emgapiv2/api/genomes.py
```diff
@@ -52,6 +53,30 @@ class GenomeFilters(BiomeFilter):
...
 
 
+def order_catalogue_genomes(qs, order):
+    if order.order in {"accession", "-accession"}:
+        prefix = "-" if order.order.startswith("-") else ""
+        return qs.order_by(f"{prefix}genome__accession")
+    return order.order_by(qs)
+
+
+def catalogue_releases_visible_to(request):
```
MGS-sails: Suggestion to rename these methods to have verb names. 
I think this would help better identify them as actions rather than properties.

emgapiv2/api/genomes.py
```diff
@@ -68,11 +93,20 @@ class GenomeController(UnauthorisedIsUnfoundController):
...
         ),
+        auth=[WebinJWTAuth(), NoAuth()],
     )
-    def get_mgnify_genome(self, accession: str):
+    def get_mgnify_genome(
+        self, request, accession: str, catalogue_id: str | None = None
+    ):
+        genomes = (
+            catalogue_genomes_visible_to(request)
+            if catalogue_id
```
MGS-sails, line 103: I think this check might be redundant, as catalogue_genomes_visible_to  will return GenomeCatalogue.public_objects anyway if the user does not have the right permissions

emgapiv2/config.py
```diff
@@ -322,6 +322,22 @@ class GenomeConfig(BaseModel):
     results_directory_root: str = "/nfs/donco/results"
     genomes_ftp_results_subpath: str = "mgnify_genomes"
 
+    ftp_genomes_root: str = "/path/to/ftp-genomes"
+
+    genome_search_project_dir: str = "/path/to/genome-search-project"
```
MGS-sails, line 327: Are these placeholder values ?

genomes/migrations/0011_finish_catalogue_release_schema.py
```diff
@@ -0,0 +1,85 @@
+import django.db.models.deletion
+from django.db import migrations, models
+from django.db.models import Q
+
+
+class Migration(migrations.Migration):
```
MGS-sails, line 6: What would happen if this migration needed to be rolled back ?

MGS-sails: I wonder if Django has something like a "down" method in it's migrations, where you could define the code that should run when a migration is rolled back

genomes/models/genome_catalogue.py
```diff
@@ -1,24 +1,60 @@
...
+        READY = "ready", "Ready to publish"
+        PUBLISHED = "published", "Published"
+        RETIRED = "retired", "Retired"
 
+    PROK = GenomeCatalogueSeries.PROK
+    EUKS = GenomeCatalogueSeries.EUKS
+    VIRS = GenomeCatalogueSeries.VIRS
+    CATALOGUE_TYPE_CHOICES = GenomeCatalogueSeries.CATALOGUE_TYPE_CHOICES
+
+    objects = models.Manager()
```
MGS-sails, line 34: could we use more explicit names for these variables, especially as "objects" is used quite frequently in other contexts

genomes/models/genome_catalogue.py
```diff
@@ -54,29 +87,143 @@ class GenomeCatalogue(WithDownloadsModel, TimeStampedModel):
...
+    @property
+    def catalogue_type(self):
+        return self.series.catalogue_type
+
+    @property
+    def biome(self):
+        return self.series.biome
+
+    @classmethod
+    def publish_ready(cls, catalogue_ids):
```
MGS-sails: I can see that catalogue_ids are the ids of catalogues the user has selected to publish. 
Could we make this var name a bit more indicative of that. for example selected_catalogue_ids.

workflows/flows/import_genomes_flow.py
```diff
@@ -149,24 +472,43 @@ def import_genomes_flow(
...
-        genome_accessions.append(genome_accession)
-    logger = get_run_logger()
-    logger.info(
-        f"Processed {len(genome_accessions)} genomes in catalogue {catalogue.name}"
+        process_genome_dir(catalogue, genome_dir)
+    get_run_logger().info(
+        f"Processed {len(genome_dirs)} genomes in catalogue {catalogue.name}"
     )
-    validate_import_summary(catalogue)
+    validate_import_summary(catalogue, expected_count=len(genome_dirs))
```
MGS-sails, line 498: Perhaps this should happen earlier on in the process. 
The run_genome_release_tasks would already have ran at this point and made changes.
So if validate_import_summary fails afterwards, the changes made by run_genome_release_tasks would still be there.

MGS-sails (changes requested):
Thanks @SandyRogers I left a few comments.
when we merge, I think the publish_is_ready should also involve updating/creating new GenomeSearchIndex records (referring to my sourmash PR).

emgapiv2/api/genomes.py
SandyRogers: I think it covers an (albeit bizarre, but known) use case where a non-public-visible genome may be in more than one CatalogueGenome relationships at a time. The bizarre case is for catalogues that are not logically distinct from each other, e.g. supersets or where genomes are moving from one biome to another.

emgapiv2/config.py
SandyRogers: yes, expectation is they should always be overridden so we use deliberately generic paths to aid debugging if a use case forgets to set them

genomes/migrations/0011_finish_catalogue_release_schema.py
SandyRogers: (Discussed offline but for posterity) yes, Django can roll back migrations. Most of the time it is automatic/implied, e.g. delete a column that the migration adds, but for RunPython migrations it needs to be defined. Here we are using `noop` because the anti-migration is too hard to encode really, and the catalogues can always be trivially delete and re-imported either before or after the migrations are applied/unapplied.

genomes/models/genome_catalogue.py
SandyRogers: Ah, this is deliberate. `objects` overrides the default manager, and typically a different name is only used in Django if the schema needs to reserve the name `objects` for a field name that only makes sense to be called that. https://docs.djangoproject.com/en/6.0/topics/db/managers/#manager-names

emgapiv2/api/genomes.py
SandyRogers: done, thanks

genomes/models/genome_catalogue.py
SandyRogers: done, thanks

workflows/flows/import_genomes_flow.py
SandyRogers: good spot. done, thanks

